PR #33, opened 2022-03-19
feat: add HttpCache service definition

kvcache:
* Add HttpCache service and add http path spec
* Add generated descriptors file. We should put this somewhere else but I don't have a better place handy atm.
* Move kotlin subfolder to kotlin-messages, restructure a little for easier development


javascript/generate_protos.sh
```diff
@@ -2,4 +2,4 @@
 set -e
 set -x
 
-PATH=node_modules/protoc-gen-ts/bin/:$PATH protoc -I=../proto --ts_out=dist cacheclient.proto controlclient.proto
+PATH=node_modules/protoc-gen-ts/bin/:$PATH protoc -I=../proto -I=/usr/local/include --ts_out=dist cacheclient.proto controlclient.proto
```
bruuuuuuuce, line 5: do we want both `-I ../proto` and ` -I /usr/local/include`?

proto/httpcache.proto
```diff
@@ -0,0 +1,25 @@
...
+package cache_client;
+
+import "cacheclient.proto";
+import "google/api/httpbody.proto";
+import "google/api/annotations.proto";
+
+service HttpCache {
+  rpc Get (_GetRequest) returns (google.api.HttpBody) {
+    option (google.api.http) = {
+      get: "/cache/get/{cache_key}"
```
mjindalmotiv8, line 15: Thinking about pre-signed URL's I am not sure if query params need to be a part of this? For pre-signed URLs I had imagined this to be "/cache/get?token={token}" where the `cache_key` was a part of the token as we need to sign the key as well

proto/httpcache.proto
```diff
@@ -0,0 +1,25 @@
...
+    };
+  }
+  rpc Set (_SetRequest) returns (_SetResponse) {
+    option (google.api.http) = {
+      post: "/cache/set/{cache_key}"
+      body: "cache_body"
+      // ?ttl_milliseconds=ttlMillis
+    };
+  }
```
virratanasangpunth, line 24: I'm assuming that envoy will just convert AuthToken/CacheName http headers into grpc headers?

proto/httpcache.proto
kvcache: Yeah I was going to do that as a second change but I can't come up with any reason not to do it now. Thanks, will update!

I think we need something along the lines of:
`/cache/get/{cache_name}/{cache_key}?token={cache_token}`

proto/httpcache.proto
kvcache: we can't use headers for the "vendable url" use case. I need to patch this path like Meera noted.

javascript/generate_protos.sh
kvcache: /usr/local/include puts protocol buffers typedefs on the find path. I think that was always the intent, no? I don't care that much, it's just a bomb for $future_dev if they add a `protobuf` typed thing into our api ever. 🤷

proto/httpcache.proto
mjindalmotiv8: Thanks!! Do we need the `cache_key` to be a part of the path.

Another question - So for non-presigned URLs where will the usual GRPC headers go (auth and cache name)

kvcache: for non-presigned HTTP interactions they will go to exactly this same path. We're not signing the URL, we're granting access via the token; so a limited token will only allow interactions to a limited key range (e.g., 1 key) and an unlimited token (e.g., our current keys) will allow interactions to the whole cache

gautamomento (approved):
This approach LGTM. I see your comment about updating the path to support tokens in the URL. I am okay either way if you do it in this PR or the next one.

kvcache:
@mjindalmotiv8 @gautamomento if I understand correctly the query parameters are implicitly able to set fields on the request bodies. If that's not how it be, I'll do a follow-up.

gautamomento (approved):

PR #34, opened 2022-03-22
feat: add create/revoke signing key to control client protos

wandaitzuchen:
Add create/~delete~revoke SigningKey to control service.
https://github.com/momentohq/cacheadmin-v2/issues/489


proto/controlclient.proto
```diff
@@ -37,3 +39,19 @@ message _ListCachesResponse {
   repeated _Cache cache = 1;
   string next_token = 2;
 }
+
+message _CreateSigningKeyRequest {
+}
+
+message _CreateSigningKeyResponse {
```
kvcache, line 46: What about expiry? Should we start out with an expiration time for these secrets? E.g., 30 days to start with? Maybe we should communicate that back to the user.

proto/controlclient.proto
```diff
@@ -37,3 +39,19 @@ message _ListCachesResponse {
...
+  string key_id = 1;
+  bytes private_key = 2;
+  int32 created_at = 3;
```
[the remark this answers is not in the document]
wandaitzuchen: `created_at`? It's just an fyi for the customer that they could use to manage the key rotation.  During a previous meeting with Wyze we talked about they themselves managing the key rotation - whenever they want to rotate the key they'll generate a new one and delete the old one using the api.

proto/controlclient.proto
wandaitzuchen: Yeah we could consider doing that too.  For now we kind of settled on having them do the key management for the private key (storing them securely, rotation, etc.).  Down the road we can add a layer of expiry on our end as well, or even change the way how key is distributed/rotated (along with our user management story?)  
@mjindalmotiv8 @danielamiao  let me know if you have any suggestions.

mjindalmotiv8: Yup - the expiry is something which I was hoping the whole user management story would guide us on so I had not included it in the design doc. I do not know for example if we will have an API which they can use to see when their private key will expire. What is the behavior of the pre-signed URLs using the private key - will they stop working? Will that cause a data plane impact?

proto/controlclient.proto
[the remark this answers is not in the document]
kvcache: Yeah but isn't it possible for them to simply call `new Date()` or `System.currentTimeMillis()` or any of a hundred other "what time is it" functions? Why should we provide "current time" from this api?

proto/controlclient.proto
kvcache: this is not a "down the road" thing - if we ever let users generate eternal signing keys then we commit to storing their public keys forever. It's more permanent even than our current eternal jwt story...

@mjindalmotiv8 If a request uses a token that is signed by an expired signature then that request will fail as not-authorized.

I think we should start out with an expiry on these tokens, I do not think that is a reasonable fast-follow.

proto/controlclient.proto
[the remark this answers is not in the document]
wandaitzuchen: Yeah that's true.  I'll remove this but instead add `expires_at` (given that we will cap the ttl at a max of say 90 days).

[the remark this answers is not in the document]
mjindalmotiv8: @wandaitzuchen - probably not related to the proto but more to dynamo where we will store the public key and id, we should store a `created_at` timestamp over there which we generate internally. This can help us in auditing or when we want to see a compromised private key generation.

proto/controlclient.proto
```diff
@@ -37,3 +39,20 @@ message _ListCachesResponse {
   repeated _Cache cache = 1;
   string next_token = 2;
 }
+
+message _CreateSigningKeyRequest {
+  int32 ttl_minutes = 1;
```
mjindalmotiv8, line 44: @kvcache - I think ttl in minutes here is ok - just wanted to get that confirmation from you

kvcache: Yeah, that seems pragmatic.

BUT

why not uint32?

mjindalmotiv8: Good catch - we should change that here and in the internal proto as well - https://github.com/momentohq/kotlin-kitchensink/pull/160/files

mjindalmotiv8 (approved):

PR #40, opened 2022-03-29
chore: fix up rest types

kvcache:
cache name and key are in the path, so let's treat them as strings
in the http path bindings.


proto/httpcache.proto
```diff
@@ -10,14 +10,14 @@ import "google/api/httpbody.proto";
...
+  string cache_name = 1;
+  string cache_key = 2;
   string token = 3;
 }
 
 message _HttpSetRequest {
-  bytes cache_name = 1;
-  bytes cache_key = 2;
+  string cache_name = 1;
+  string cache_key = 2;
```
virratanasangpunth, line 20: What's envoy doing here? Is it base64 encoding?

Do we require that these chars `/` `=` and `+` are percent-encoded?

kvcache: With strings this will only support url path-safe cache names and cache keys. Values are in the body so it _should_ support arbitrary values. Still working on making sure about that part.

virratanasangpunth: Oh I see there's no base64 encoding here for cache_name/cache_key.

virratanasangpunth (approved):

proto/httpcache.proto
kvcache: Right, thanks!

PR #43, opened 2022-04-04
chore: also update the HttpSetRequest to use uint64

tylerburdsall:
Looks like we also need to change this type to `uint64` so that our HttpClient is also compatible


kvcache (approved):
yup

virratanasangpunth (approved):

PR #44, opened 2022-04-08
chore: change HttpSetRequest cache_body to an HttpBody

tylerburdsall:


kvcache (approved):
Note: This breaks `/cache/set/` for any HTTP client that supports alpn H2 (e.g., curl). For curl specifically you can work around with `--http1.1` but not all clients are expected to have trivial workarounds.

PR #45, opened 2022-04-18
feat: add protos for ListSigningKeys

tylerburdsall:
First step for `ListSigningKeys` support


proto/controlclient.proto
```diff
@@ -55,3 +56,17 @@ message _RevokeSigningKeyRequest {
 
 message _RevokeSigningKeyResponse {
 }
+
+message _SigningKey {
+  string key = 1;
```
virratanasangpunth, line 61: `key_id` you mean?

tylerburdsall: Yup, thank you for catching this!

proto/controlclient.proto
```diff
@@ -55,3 +56,17 @@ message _RevokeSigningKeyRequest {
 
 message _RevokeSigningKeyResponse {
 }
+
+message _SigningKey {
+  string key_id = 1;
+  uint64 expires_at = 2;
```
kvcache, line 62: Nit: I fully realize nothing else in here is commented. Would you consider putting comments on the fields in these new messages?

proto/controlclient.proto
```diff
@@ -55,3 +56,17 @@ message _RevokeSigningKeyRequest {
...
+}
+
+message _ListSigningKeysRequest {
+  string next_token = 1;
```
kvcache, line 66: I remain saddened that we don't use streaming responses for these.

kvcache (commented):
nits no blockers

proto/controlclient.proto
tylerburdsall: Sure thing!

proto/controlclient.proto
tylerburdsall: > I remain saddened that we don't use streaming responses for these.

Yeah, I would like that too and it seems like we aren't using pagination atm. I wanted to keep this here to at least maintain similar patterns

kvcache (approved):

PR #46, opened 2022-04-21
chore: add necessary lines to easily support updates to Go SDK

tylerburdsall:
Added necessary lines to make it easier to copy over to https://github.com/momentohq/client-sdk-go when making proto changes

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -1,5 +1,6 @@
 syntax = "proto3";
 
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
 option java_multiple_files = true;
 option java_package = "grpc.cache_client";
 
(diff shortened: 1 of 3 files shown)


poppoerika (commented):
Thank you for adding go package option!
Could you also add README or something to show which command we can use to generate files from the protos in the `proto` directory? 
I'd assume this won't be the last time to generate files from new protos.

tylerburdsall:
> Thank you for adding go package option! Could you also add README or something to show which command we can use to generate files from the protos in the `proto` directory? I'd assume this won't be the last time to generate files from new protos.

Sure thing @poppoerika  I will be making those changes in the `client-sdk-go` package since it appears as though they need to be generated from that package instead of this repo

poppoerika (approved):

PR #48, opened 2022-04-25
[DO NOT MERGE] chore: make compatible with .NET Framework 4.5.2

### LINKED ISSUE #47: [csharp] Fork this repo and make it compatible with `netstandard1.2`
Original ticket: https://github.com/momentohq/client-sdk-csharp/issues/48
In order to use proto generated files in our C# SDK, we need to update the project file to include `netstandard1.2` and use `Grpc.Core` instead of`Grpc.Net.Client`.

poppoerika:
We don't want to merge this pr since we want to have a separate branch `clientProtosDotNetFramework4.5.2` for older .NET Frameworks without affecting users with more recent .NET Frameworks/Cores.
Relevant conversations: https://github.com/momentohq/client-sdk-csharp/pull/49
Relevant ticket: https://github.com/momentohq/client-sdk-csharp/issues/48

- Use `Grpc.Core` instead of `Grpc.Net.Client` to support .NET Framework 4.5.2
- I think we need to manually generate files from the proto in a Windows Server 2012 R2 instance and upload it to the artifactory since Windows Server 2022/2019 are the only GitHub-Hosted runners which don't support .NET Framework less than 4.7.2/4.8.
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#net-framework
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#net-framework

Closes #47

### DIFF
--- a/csharp/Momento.csproj
+++ b/csharp/Momento.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
+    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
   </PropertyGroup>
 
 <ItemGroup>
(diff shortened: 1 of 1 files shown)


eaddingtonwhite:
@poppoerika can we close this based off your findings on other ticket on GRPC support for older versions of .net?

danielamiao (commented):
Echoing Ellery's comment, we should close off this PR since it seems Grpc.Core won't work for us either.

PR #49, opened 2022-04-26
feat: add delete api to scs service

bruuuuuuuce:


proto/cacheclient.proto
```diff
@@ -31,6 +32,14 @@ message _GetResponse {
   string message = 3;
 }
 
+message _DeleteRequest {
+  bytes cache_key = 1;
+}
+
+message _DeleteResponse {
+  ECacheResult result = 1;
```
kvcache, line 40: This field is unnecessary. Delete invocations either succeed (grpc OK) or fail/error (anything other than grpc OK)

proto/cacheclient.proto
```diff
@@ -31,6 +32,12 @@ message _GetResponse {
   string message = 3;
 }
 
+message _DeleteRequest {}
```
gautamomento, line 35: I am just curious here.

I would have thought the delete here means that we want to remove an entry from the cache. Is there a `bytes cache_key = 1` needed here?

bruuuuuuuce: oops yes I removed the wrong one 😅 . good catch

kvcache (approved):

gautamomento (approved):

PR #51, opened 2022-05-09
feat: Add proto declarations for dictionary.

### LINKED ISSUE #50: Add public protos for new HashMap apis

malandis:
Closes #50 

Add the rpc method declarations for DictionaryGet, DictionaryGetAll DictionarySet and the associated request and response messages. Terminology:
- dict, dictionary: a data structure that supports key-based indexing
- dictionary name: the name of a dictionary stored in momento (ie the cache key for the dictionary)
- dictionary key: a key to index into a dictionary
- dictionary value: the value corresponding to a key in a particular dictionary (ie `dictionary_value = dictionary_name.get(dictionary_key)`)

The following summarizes the methods:

- DictionaryGet: request the dictionary value for the dictionary key `dictionary_key` from a dictionary `dictionary_name`. The response message is identical to that of `_GetResponse`
- DictionaryGetAll: request the entire dictionary `dictionary_name` from the cache. The response message is parallel to `_DictionaryGetResponse` except the value to return is a list `DictionaryKeyValuePair`*.
- DictionarySet: upsert a dictionary `dictionary_name` with the keys and values from `dictionary_body`*. Consistent with the rest of the API, keys and values are bytes. The response message is identical to that of `_SetResponse`.

* Note protobuf does have a `map` type, which would be helpful here. Unfortunately it does not permit bytes as keys or values. Hence to represent `map<bytes, bytes>`, we use a repetition of `DictionaryKeyValuePair` messages, each of which defines a `key` (`bytes`), and `value` (`bytes`).

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -20,9 +20,9 @@ service Scs {
   rpc Get (_GetRequest) returns (_GetResponse) {}
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
-  rpc HashGet (_HashGetRequest) returns (_HashGetResponse) {}
-  rpc HashGetAll (_HashGetAllRequest) returns (_HashGetAllResponse) {}
-  rpc HashSet (_HashSetRequest) returns (_HashSetResponse) {}
+  rpc DictGet (_DictGetRequest) returns (_DictGetResponse) {}
+  rpc DictGetAll (_DictGetAllRequest) returns (_DictGetAllResponse) {}
+  rpc DictSet (_DictSetRequest) returns (_DictSetResponse) {}
 }
 
 message _GetRequest {
@@ -52,38 +52,38 @@ message _SetResponse {
   string message = 2;
 }
 
-message _HashGetRequest {
-  bytes cache_hash_name = 1;
-  bytes cache_hash_key = 2;
+message _DictGetRequest {
+  bytes dict_name = 1;
+  bytes dict_key = 2;
 }
 
-message _HashGetResponse {
+message _DictGetResponse {
   ECacheResult result = 1;
   bytes cache_body = 2;
   string message = 3;
 }
 
-message _HashGetAllRequest {
-  bytes cache_hash_name = 1;
+message _DictGetAllRequest {
+  bytes dict_name = 1;
 }
 
-message HashKeyValuePair {
+message DictKeyValuePair {
   bytes key = 1;
   bytes value = 2;
 }
 
-message _HashGetAllResponse {
+message _DictGetAllResponse {
   ECacheResult result = 1;
-  repeated HashKeyValuePair cache_body = 2;
+  repeated DictKeyValuePair dict_body = 2;
   string message = 3;
 }
 
-message _HashSetRequest {
-  bytes cache_hash_name = 1;
-  repeated HashKeyValuePair mapping = 2;
+message _DictSetRequest {
+  bytes dict_name = 1;
+  repeated DictKeyValuePair dict_body = 2;
 }
 
-message _HashSetResponse {
+message _DictSetResponse {
   ECacheResult result = 1;
   string message = 2;
 }
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -67,15 +67,20 @@ message _HashGetAllRequest {
   bytes cache_hash_name = 1;
 }
 
+message HashKeyValuePair {
+  bytes key = 1;
+  bytes value = 2;
+}
+
 message _HashGetAllResponse {
   ECacheResult result = 1;
-  map<bytes, bytes> cache_body = 2;
+  repeated HashKeyValuePair cache_body = 2;
   string message = 3;
 }
 
 message _HashSetRequest {
   bytes cache_hash_name = 1;
-  map<bytes, bytes> mapping = 2;
+  repeated HashKeyValuePair mapping = 2;
 }
 
 message _HashSetResponse {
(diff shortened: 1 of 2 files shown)


proto/cacheclient.proto
```diff
@@ -48,3 +51,39 @@ message _SetResponse {
   ECacheResult result = 1;
   string message = 2;
 }
+
+message _HashGetRequest {
+  bytes cache_hash_name = 1;
+  bytes cache_hash_key = 2;
```
kvcache, line 57: Maybe just hash_name and hash_key? They're going to be handled quite differently so maybe it would be confusing to overlap the names.

proto/cacheclient.proto
```diff
@@ -48,3 +51,34 @@ message _SetResponse {
   ECacheResult result = 1;
   string message = 2;
 }
+
+message _HashGetRequest {
```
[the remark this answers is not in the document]
cprice404: to paraphrase kenny's point, he pointed out that there are many different names for this data structure in different tools/languages.  hash, associative array, map, object, dictionary.  And we shouldn't consider ourselves locked into one noun that our competitors use.

I think "map" and "object" are too overloaded.  I think "associative array" is too wordy.

I like "dict" and "hash" the best.  Personally I would tie-break between those two in favor of "dict" simply because the word "hash" is also kind of overloaded in programming.  "dict" doesn't really have any other meanings.

So I'd probably vote for `Dict*`, but `Hash*` would be my second choice.

proto/cacheclient.proto
```diff
@@ -48,3 +51,39 @@ message _SetResponse {
...
+}
+
+message HashKeyValuePair {
+  bytes key = 1;
+  bytes value = 2;
+}
+
+message _HashGetAllResponse {
+  ECacheResult result = 1;
+  repeated HashKeyValuePair cache_body = 2;
```
kvcache, line 77: I think this should be hash_body

cprice404 (commented):
🙈 other than naming considerations.

proto/cacheclient.proto
```diff
@@ -48,3 +51,39 @@ message _SetResponse {
...
+  string message = 3;
+}
+
+message _HashSetRequest {
+  bytes cache_hash_name = 1;
+  repeated HashKeyValuePair mapping = 2;
```
kvcache, line 83: With this type you'll also need to provide a hash_delete() api with a key list

kvcache (commented):
Few thoughts mostly looks fine

proto/cacheclient.proto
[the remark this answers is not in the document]
kvcache: Thanks Chris. If we're open to this I prefer Dict too because we also say hash as a verb. Dicts are not verbs though and imo are a better naming candidate for this new noun in our api surface.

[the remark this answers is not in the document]
malandis: > did we bottom out on `Hash` as the noun here? Kenny had left some comments somewhere that I thought were interesting on this topic.

No. I wanted to field this for broader discussion.

I agree with Kenny's points that "hash" is not the best. A hash could be a hash table, a hash function, the image of a hash function, a hashmap, ... etc. Also hashing is a particular way to implement a dictionary. I don't see anyone clamoring for the name "RedBlackTreeMap".

`map` could be the `map` higher-order function. `object` is vague and `associative array` is too long like you said. I think Dict* could be good.

mjindalmotiv8:
I am fine with this once you have addressed the changes suggested by Kenny and Chris!! 

Thanks for the changes!!

proto/cacheclient.proto
malandis: Depending how much granularity we want:
(1) delete the entire hash from momento (no key list required)
(2) delete a list of keys

proto/cacheclient.proto
malandis: Using the dictionary terminology, would a better name for this just be "dictionary"?

proto/cacheclient.proto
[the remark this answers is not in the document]
malandis: Renamed Hash->Dict.

proto/cacheclient.proto
malandis: Changed to "dict_name" and "dict_key"

proto/cacheclient.proto
kvcache: You'll have _name in other places so for consistency maybe use _name and _body

malandis:
@cprice404 @kvcache @mjindalmotiv8 : regarding `DictSet` -- in the current implementation we pass a repetition of `DictKeyValuePair`s. Are there any pros/cons of doing this vs making the method streaming request on a single `DictKeyValuePair`?

proto/cacheclient.proto
kvcache: Deleting the entire hash will prooooobably be done via the scs delete api. I think you're looking at needing (2).

proto/cacheclient.proto
malandis: Done

kvcache:
Thanks for the rename. I prefer Dict over Hash for clarity but tbh Dictionary sounds even better to my ears. It's more compatible with folks who don't enunciate all consonants and people can still say dict if they want to without ambiguity. Personally I like to steer clear of words that sound like "dick" in api naming and Dictionary is not an onerous thing to type.

proto/cacheclient.proto
malandis: I'll leave this for a separate PR. Current ask if for Get, Set, and GetAll.

cprice404 (dismissed):
thanks, i like this better than Hash personally :)

malandis:
I bumped the rust protobuf dep to resolve the failing rust build.

eaddingtonwhite (approved):
This is looking good to me @malandis nice job! I like where we landed on names as well thanks for going on that journey :). Maybe let few others who were on this give 👍 as well but +1 from me on 🚢 this.

mjindalmotiv8 (approved):

malandis:
Re-opening this for discussion as this may have been overlooked. I want to make clear we punted on TTL for dictionaries. This is why the DictionarySetRequest message does not have a ttl_milliseconds field like SetRequest does. Any objections?

PR #52, opened 2022-05-13
feat: add auth proto for login

kvcache:
Includes the Auth service stub with the inaugural Login api.

I've @'d a wide audience for FYI's. @rlinehan please do look!


proto/momento/auth.proto
```diff
@@ -0,0 +1,42 @@
...
+option java_package = "momento.auth";
+
+package momento.auth;
+
+service Auth {
+  rpc Login (_LoginRequest) returns (stream _LoginResponse) {}
+}
+
+message _LoginRequest {
+}
```
eaddingtonwhite, line 13: So this would be autheticated request with our JWT used to get sso login url? Just seems bit funny to me to have an authenticated request called _LoginRequest. Maybe more clear to call `SsoLogin`/ `_SsoLoginRequest`?

proto/momento/auth.proto
```diff
@@ -0,0 +1,42 @@
...
+
+message _LoginResponse {
+  oneof state {
+    DirectBrowser direct_browser = 1;
+    LoggedIn logged_in = 2;
+    Message message = 3;
+    Error error = 4;
```
eaddingtonwhite, line 20: How do I get into the DirectBrowser or Message state during a login? Are those intermediary states on users path  to LoggedIn status? Just confused how I get into either of those states here. LoggedIn or Error seems clear to me.

proto/momento/auth.proto
kvcache: > So this would be autheticated request with our JWT used to get sso login url?

No. This is untrusted callers asking to log in.

> Just seems bit funny to me to have an authenticated request called _LoginRequest.

It would be yes, but no - this is how unauthenticated users become authenticated users.

HMU on slack if you'd like to have a deeper conversation on the topic!

proto/momento/auth.proto
kvcache: These are tools for SDK writers to implement the login flow. I will consume these states in `momento`. Their semantics are briefly described on the respective messages' definitions.

Each of these messages is a command from the service telling the login process on the client machine what to do.
* DirectBrowser tells the client machine to... direct a browser to the url.
* LoggedIn tells the client machine that the user has logged in (and gives it the session token).
* Message delivers a message destined for the user (implemented, e.g., via printf)
* Error tells the client that something went wrong, and a description of what.
* The end of the grpc `stream` response indicates that the login is done. LoggedIn and Error happen to be the 2 intended final state messages.

I don't think you're asking for an exhaustive listing of all defined oauth2 login dispositions but I'm not sure what you're looking for otherwise. I can point you to the server-side workflow implementation that is already checked in if you like?

eaddingtonwhite: I see this helps! I think was just just confused on how this was used as stream back and forth. You descriptions helped me visualize more and makes sense 👍 .

proto/momento/auth.proto
eaddingtonwhite: Ok that makes sense to me then didn't know if this would be started totally unauthenticated from cli I think if thats case then this name is good 👍

eaddingtonwhite (approved):
LGTM seems pretty straight forward! Excited to see this working e2e.

proto/momento/auth.proto
kvcache: Yeah, today we don't have any unauthenticated -> authenticated API edge beyond an email. This will bring users in via normal identity provider idioms and establish a repeatable identity securely. It's going to be real neato.

proto/momento/auth.proto
kvcache: cool - yeah, login is actually is a one-way stream. The IDP flow is out-of-band and happens (at least in this iteration) via the DirectBrowser state though we might have other flows for special cases in the future.

proto/momento/auth.proto
```diff
@@ -0,0 +1,42 @@
+syntax = "proto3";
```
rlinehan, line 1: Since this is a public repo and this is still in flux, do you think it would make sense to leave a comment at the top? Not entirely sure how much attention this code is getting, but it is going to be a minute since this is all implemented?

rlinehan (approved):
This looks good to me and matches my understanding of what we're going to need here!

proto/momento/auth.proto
```diff
@@ -0,0 +1,42 @@
...
+
+option java_multiple_files = true;
+option java_package = "momento.auth";
+
+package momento.auth;
+
+service Auth {
+  rpc Login (_LoginRequest) returns (stream _LoginResponse) {}
```
cprice404, line 9: This isn't entirely related to this PR, but the way this is trending it seems as though this will be a cellular login endpoint.  Have we had any discussion about how this would relate to / interact with any global state that we might want to store about auth?  Or are we planning for all of our auth state to be entirely cellular?

kvcache: will slack you

proto/momento/auth.proto
kvcache: If someone tries to use it they'll get a fine message back from grpc like `Error invoking method "momento.Auth/Login": target server does not expose service "momento.Auth"`. We don't currently offer support for people writing their own clients & the Auth service will soon be a fully supported service. I don't really want to add disclaimers to it <3

rlinehan: Cool that makes sense. Just wanted to check.

kvcache: Yeah, it's a worthwhile thing to talk about! In general I'm okay with putting production-destined api definitions in production locations. This one is pretty well-understood and has a pattern we're laying down right now so it shouldn't be too much of an upset to have the proto definitions ahead of the implementation's availability.

PR #53, opened 2022-05-13
feat: Adding support for PUTs

mjindalmotiv8:
Unfortunately need to declare this as a separate RPC because only oneof PUT or POST is allowed. Still needs to be verified with cacheservice which I will do shortly.


tylerburdsall (approved):
Approving if the tests work out

kvcache (approved):
mmhmm

PR #59, opened 2022-05-17
feat: generate grpc services for rust

kvcache:
* replace dependency on environment protoc with tonic-build.
* simplify build environment due to generating protos differently.


bruuuuuuuce (approved):
:shipit:

PR #60, opened 2022-06-01
chore: add validity time to logged in message

kvcache:
Will enable some more convenient client-side heuristics around
session refreshing.

### DIFF
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -23,6 +23,15 @@ message _LoginResponse {
   // Terminal state, login success
   message LoggedIn {
     string session_token = 1;
+
+    // How many the token was valid for when issued.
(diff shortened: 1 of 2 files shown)


proto/auth.proto
```diff
@@ -23,6 +23,15 @@ message _LoginResponse {
   // Terminal state, login success
   message LoggedIn {
     string session_token = 1;
+
+    // How many the token was valid for when issued.
```
rlinehan, line 27:
```suggestion
    // How many seconds the token was valid for when issued.
```

rlinehan (dismissed):
looks good, one missing word in the comment

rlinehan (dismissed):

rlinehan (approved):

PR #61, opened 2022-06-03
feat: Support variadic keys in DictionaryGet

malandis:
Change DictionaryGetRequest to support any number of keys in the
request. Similarly modify DictionaryGetResponse to return multiple
values, each with the cache status (HIT, MISS, etc) and cache body.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -54,33 +54,36 @@ message _SetResponse {
 
 message _DictionaryGetRequest {
   bytes dictionary_name = 1;
-  bytes dictionary_key = 2;
+  repeated bytes dictionary_keys = 2;
 }
 
 message _DictionaryGetResponse {
-  ECacheResult result = 1;
-  bytes cache_body = 2;
-  string message = 3;
+  message _DictionaryGetResponsePart {
+    ECacheResult result = 1;
+    bytes cache_body = 2;
+  }
+  repeated _DictionaryGetResponsePart dictionary_body = 1;
+  string message = 2;
 }
 
 message _DictionaryGetAllRequest {
   bytes dictionary_name = 1;
 }
 
-message DictionaryKeyValuePair {
+message _DictionaryKeyValuePair {
   bytes key = 1;
   bytes value = 2;
 }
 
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -54,33 +54,36 @@ message _SetResponse {
...
 
 message _DictionaryGetResponse {
-  ECacheResult result = 1;
-  bytes cache_body = 2;
-  string message = 3;
+  message _DictionaryGetResponsePart {
+    ECacheResult result = 1;
+    bytes cache_body = 2;
+  }
+  repeated _DictionaryGetResponsePart dictionary_body = 1;
```
cprice404, line 65: just out of curiosity, was this a customer ask?  or does this exist in other cache engine APIs?  can follow up offline

proto/cacheclient.proto
```diff
@@ -54,33 +54,36 @@ message _SetResponse {
 
 message _DictionaryGetRequest {
   bytes dictionary_name = 1;
-  bytes dictionary_key = 2;
+  repeated bytes dictionary_keys = 2;
```
cprice404, line 57: re-using the same numeric slot in the proto with a different type breaks backward compat.

In this case I think it's the right thing to do, but just pointing it out for future consideration when we are considering changes to non-incubating APIs.

(If we have a canary that is testing this stuff I would imagine it will flutter a bit during rollout.)

cprice404 (approved):
🙈

proto/cacheclient.proto
malandis: Understood. I was aggressive here since this wasn't hooked up to the backend yet.

cprice404: cool yep i think that's the right call for now.

kvcache: Thank you for calling this out Chris. I agree this is the right thing to do since it is not hooked up yet.

proto/cacheclient.proto
malandis: I followed up offline with a link to the conversation

cprice404: thanks!

PR #63, opened 2022-06-21
feat: Add exists service method and request/response messages.

malandis:
`Exists` takes a list of keys as input and returns a list of `bool`s. `results[i]` is `True` if `keys[i]` exists, else `False`.

We decided on returning a *list* as opposed to the count of items. By doing this we can support rich iteration use cases on the client side. For example:

```python
response = client.exists("my-cache", *keys)
if not response.all():
   log.info(f"Expected {len(keys)} to exist but got {response.num_exists()}")
   for missing_key in response.missing_keys():
       # Do things
```

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -47,6 +47,7 @@ message _ExistsRequest {
 }
 
 message _ExistsResponse {
+  // Order and count of results matches the request list.
   repeated bool results = 1;
   string message = 2;
 }
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -20,6 +20,7 @@ service Scs {
   rpc Get (_GetRequest) returns (_GetResponse) {}
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
+  rpc Exists (_ExistsRequest) returns (_ExistsREsponse) {}
   rpc DictionaryGet (_DictionaryGetRequest) returns (_DictionaryGetResponse) {}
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -41,6 +42,15 @@ message _DeleteRequest {
 
 message _DeleteResponse {}
 
+message _ExistsRequest {
+  repeated bytes keys = 1;
+}
+
+message _ExistsResponse {
+  repeated bool results = 1;
+  string message = 2;
```
kvcache, line 51: can probably leave this off?

malandis: The `message` field?

proto/cacheclient.proto
```diff
@@ -41,6 +42,15 @@ message _DeleteRequest {
 
 message _DeleteResponse {}
 
+message _ExistsRequest {
+  repeated bytes keys = 1;
+}
+
+message _ExistsResponse {
+  repeated bool results = 1;
```
kvcache, line 50:
```suggestion
  // Order and count of results matches the request list.
  repeated bool results = 1;
```

proto/cacheclient.proto
kvcache: yes

mjindalmotiv8:
Hey - I am a bit behind on the details here hence my questions
1)  Does exists mean that we need to support for all key-value pairs and advanced data structures? 
2) What is the reasoning behind list of keys and not single key? Is it because Redis/memcache have similar APIs? Or have customers asked for this explicitly? Is there a limit on how many keys can be queried for?

malandis:
Closing for now. Will revisit when a priority.

PR #65, opened 2022-07-07
chore: create separate gradle project for producing java proto artifacts

cprice404:
Currently, we have a single jar artifact in the "kotlin-messages"
gradle project which bundles both java and kotlin proto artifacts.
We are also publishing this to our internal codeartifact repo.

This commit is the first of a series to clean this up and get these
published to maven central so that users can build our OSS projects
and we can get rid of the submodules on the Java SDK.

This commit introduces a new java gradle project that builds a jar
that contains only the java proto artifacts.  Subsequent commits
will include:

* Publishing this java protos jar artifact to maven central.
* Removing the java protos from the kotlin jar.
* Removing the codeartifact stuff from the kotlin jar and publishing
  that jar to maven central.
* Updating the java SDK to get rid of the submodules and consume
  the jar from maven central.
* Updating any places in the service code where we are referring
  to the kotlin artifact, to use the maven central coordinates
  instead.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -34,6 +34,34 @@ jobs:
           popd
         shell: bash
 
+  build_java:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get current time
+        uses: gerred/actions/current-time@master
+        id: current-time
+
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Install protoc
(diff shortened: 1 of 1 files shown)


kvcache:
Left a comment on the other one

cprice404:
@kvcache said:

> This looks good. Did you intend to skip the .github pr workflow though? Seems to me like that one wouldn't need any publishing path worked out

No I didn't intend to skip the PR workflow.  Can do it in next PR, or this one if it isn't merged by the time I have bandwidth for it.

cprice404:
@kvcache now has CI for java lib

kvcache (approved):
lgtm thanks

PR #67, opened 2022-07-21
feat: add protos for set data structure

anp13:


proto/cacheclient.proto
```diff
@@ -108,3 +111,58 @@ message _DictionaryDeleteRequest {
...
+  }
+
+  bytes set_name = 1;
+  oneof difference {
+    _Minuend minuend = 2;
+    _Subtrahend subtrahend = 3;
+  }
+}
+
+message _SetDifferenceResponse {}
```
anp13, line 168: @kvcache @malandis I know we decided that for read apis we would return the information if the set was found or missing and return the data if the set was found but for write apis which depend on the set to be present eg: SetDifference what should we return in the response? I think we should throw an exception here. And for the case where we want to delete the entire set (the set difference identity case) we will have to make it a no-op to be consistent with the dictionary delete all case. wdut?

kvcache: interesting. Looking forward to Mike's take.

kvcache (approved):
looks great, thanks

PR #70, opened 2022-08-01
feat: apply name proposal to DictionaryGetAll

malandis:
In line with the name proposal, this renames DictionaryGetAll* to
DictionaryFetch*.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -21,7 +21,7 @@ service Scs {
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
   rpc DictionaryGet (_DictionaryGetRequest) returns (_DictionaryGetResponse) {}
-  rpc DictionaryGetAll (_DictionaryGetAllRequest) returns (_DictionaryGetAllResponse) {}
+  rpc DictionaryFetch (_DictionaryFetchRequest) returns (_DictionaryFetchResponse) {}
   rpc DictionarySet (_DictionarySetRequest) returns (_DictionarySetResponse) {}
   rpc DictionaryDelete (_DictionaryDeleteRequest) returns (_DictionaryDeleteResponse) {}
   rpc SetFetch (_SetFetchRequest) returns (_SetFetchResponse) {}
@@ -70,7 +70,7 @@ message _DictionaryGetResponse {
   string message = 2;
 }
 
-message _DictionaryGetAllRequest {
+message _DictionaryFetchRequest {
   bytes dictionary_name = 1;
 }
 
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -70,7 +70,7 @@ message _DictionaryGetResponse {
   string message = 2;
 }
 
-message _DictionaryGetAllRequest {
+message _DictionaryFetchRequest {
```
anp13, line 73: We also need to update the `DictionaryGetResponse` and `DictionaryFetchResponse` to the `Found/Missing` format like we did for `SetFetch` and also remove the `ECacheResult` from all the dictionary api responses.

proto/cacheclient.proto
```diff
@@ -58,53 +58,66 @@ message _SetResponse {
...
+  message _Found {
+    repeated _DictionaryGetResponsePart items = 1;
+  }
+
+  message _Missing {}
+
+  oneof dictionary {
+    _Found found = 1;
+    _Missing missing = 2;
+  }
```
malandis, line 79: I applied the `_Found` (`_Missing`) pattern to the dictionary overall but not the unary items within.

proto/cacheclient.proto
malandis: Added `Found/Missing` in caf8c1a.

proto/cacheclient.proto
```diff
@@ -58,53 +58,66 @@ message _SetResponse {
...
 message _DictionaryDeleteRequest {
   message Some {
-    repeated bytes keys = 1;
+    repeated bytes fields = 1;
   }
   message All { }
 
   bytes dictionary_name = 1;
-  oneof delete_keys {
+  oneof delete_fields {
```
anp13, line 120: nit: we can call this `delete` to be consistent with the set and list apis. We use `difference` in SetDifferenceRequest and for list apis also I have used `erase` in ListEraseRequest and `remove` in ListRemoveRequest. 

https://github.com/momentohq/client_protos/pull/71/files

anp13 (commented):
Looks good just one comment. 

Do you think we should change `Found` to `Ok` especially for update api responses? `Found` seems weird in api responses which do not return anything.

proto/cacheclient.proto
malandis: Fixed in a332286

malandis:
> Looks good just one comment.
> 
> Do you think we should change `Found` to `Ok` especially for update api responses? `Found` seems weird in api responses which do not return anything.

Do you mean for the `Set` APIs?

anp13:
> > Looks good just one comment.
> > Do you think we should change `Found` to `Ok` especially for update api responses? `Found` seems weird in api responses which do not return anything.
> 
> Do you mean for the `Set` APIs?

Yes. It only applies to the `SetDifference` API I think. I think it is ok to keep it as `Found` for now since we are not using it in the SDK yet.

anp13 (approved):

PR #71, opened 2022-08-02
feat: add list api protos

anp13:


kvcache (approved):
lgtm

proto/cacheclient.proto
```diff
@@ -175,3 +185,105 @@ message _SetDifferenceResponse {
...
+message _ListPushBackResponse {}
+
+message _ListPopFrontRequest {
+  bytes list_name = 1;
+}
+
+message _ListPopFrontResponse {
+  message _Found {
+    bytes front = 1;
+  }
```
malandis, line 216: Can we easily change this to support a request to pop `count` items from a list `list_name`?

proto/cacheclient.proto
```diff
@@ -175,3 +185,105 @@ message _SetDifferenceResponse {
...
+
+message _ListPopBackRequest {
+  bytes list_name = 1;
+}
+
+message _ListPopBackResponse {
+  message _Found {
+    bytes back = 1;
+  }
+
```
malandis, line 234: Same question as for ``ListPopFront`

proto/cacheclient.proto
anp13: Good callout. @kvcache we should make the change on the blobdb server as well.

anp13: @malandis going to leave this as is for now, should be easy to add this functionality later.

malandis (approved):

PR #72, opened 2022-08-03
chore: add metadata to csproj file for NuGet

pgautier404:
Add extended metadata for display on the public NuGet package repository.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -165,21 +165,7 @@ jobs:
 
       - uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: "5.0.x"
-
-      - uses: nuget/setup-nuget@v1
-        with:
-          nuget-version: "5.x"
-
-      - name: Configure Artifactory publish credentials
-        run: |
-          set -e
-          set -x
-          pushd csharp
-            nuget sources Add -Name Artifactory -Source https://momento.jfrog.io/artifactory/api/nuget/nuget-local -username deploy-user -password ${{ secrets.PUBLIC_ARTIFACTORY_DEPLOY_USER_MAGIC_PASSTOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -188,10 +174,14 @@ jobs:
             VERSION="${{needs.release.outputs.version}}"
             echo "version: ${VERSION}"
             dotnet build --configuration Release
-            nuget pack -Version ${VERSION} -Properties Configuration=Release
-            nuget push Momento.${VERSION}.nupkg -Source Artifactory -Verbosity detailed
+            dotnet pack -property:Version=${VERSION} --configuration Release
+            pwd
```
malandis, line 178: Is this line just for debug output?

csharp/Momento.csproj
```diff
@@ -2,6 +2,16 @@
...
+    <PackageId>Momento.Protos</PackageId>
+    <Authors>Momento</Authors>
+    <Company>Momento Inc</Company>
+    <Description>
+      Generated gRPC and Protocol Buffers types for Momento clients
+    </Description>
+    <PackageTags>caching, cache, serverless, key value, simple caching service</PackageTags>
+    <Copyright>Copyright (c) Momento Inc 2022</Copyright>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <ProjectUrl>https://github.com/momentohq/client_protos/tree/main/csharp</ProjectUrl>
```
malandis, line 14: The URL should just be to /client_protos/

malandis (changes requested):

.github/workflows/push.yml
pgautier404: Ah, yes, will remove. Thanks.

csharp/Momento.csproj
pgautier404: I modeled that after the pypi info, but I tend to agree with you. I've changed it.

malandis (approved):

PR #76, opened 2022-08-19
feat: javascript-webtext

kvcache:
Add grpc-web-text generated types so momento grpc can be used by
browsers. https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md

Tested with a local package.json dependency from a create-react-app.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -33,6 +33,35 @@ jobs:
             npm run build
           popd
         shell: bash
+  
+  build_javascript-web:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get current time
+        uses: gerred/actions/current-time@master
+        id: current-time
+
+      - uses: actions/checkout@v2
+
+      - name: Install Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+
+      - name: Install protoc
+        run: ./install_protoc.sh
+
+      - name: Build package
+        run: |
+          set -e
+          set -x
+          export PATH=`pwd`/protoc/bin/:$PATH
+          pushd javascript
+            npm ci
+            npm run build
+          popd
+        shell: bash
 
   build_java:
     runs-on: ubuntu-latest
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -34,7 +34,7 @@ jobs:
           popd
         shell: bash
   
-  build_javascript-web:
+  build_javascriptweb:
     runs-on: ubuntu-latest
 
     steps:
@@ -57,7 +57,7 @@ jobs:
           set -e
           set -x
           export PATH=`pwd`/protoc/bin/:$PATH
-          pushd javascript
+          pushd javascript-web
             npm ci
             npm run build
           popd
--- a/javascript-web/README.md
+++ b/javascript-web/README.md
@@ -1,0 +1,40 @@
+# Usage
+### package.json
+```
+dependencies: {
+  "@gomomento/generated-types-webtext": "$current_version",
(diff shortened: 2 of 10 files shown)


bruuuuuuuce (approved):
:shipit: sweet!!!!

javascript-web/README.md
```diff
@@ -0,0 +1,40 @@
...
+dependencies: {
+  "@gomomento/generated-types-webtext": "$current_version",
+  "grpc-web": "1.3.1",
+  "google-protobuf": "3.21.0"
+}
+```
+
+### Basic usage shape
+```typescript
+const authToken = getClientAuthToken_never_harcode_server_api_keys();
```
bruuuuuuuce, line 13: it looks like eventually we will have to build a separate web client?

kvcache: Going to want to make separate distributions in order to have the appropriate trim dependencies.

Instead of copy pasting the whole thing we might end up making the momento client rely on an internal protocol implementation, which we could either bundle into assembly-packages (better) or require users to choose dependencies (worse).

.github/workflows/pull_request.yml
```diff
@@ -33,6 +33,35 @@ jobs:
...
+
+      - name: Install protoc
+        run: ./install_protoc.sh
+
+      - name: Build package
+        run: |
+          set -e
+          set -x
+          export PATH=`pwd`/protoc/bin/:$PATH
+          pushd javascript
```
cprice404, line 60: should this be `javascript-web`?

javascript-web/README.md
```diff
@@ -0,0 +1,40 @@
+# Usage
+### package.json
+```
+dependencies: {
+  "@gomomento/generated-types-webtext": "$current_version",
```
cprice404, line 5: Are you sure you are happy with this module name for npmjs publishing?  I'm not challenging it, just want to ensure we are happy with it since it's a one-way-door.

javascript-web/README.md
```diff
@@ -0,0 +1,40 @@
...
+  "grpc-web": "1.3.1",
+  "google-protobuf": "3.21.0"
+}
+```
+
+### Basic usage shape
+```typescript
+const authToken = getClientAuthToken_never_harcode_server_api_keys();
+const cacheName = 'your_cool_cache';
+const momento = new cache.ScsClient('https://cache.cell-us-east-1-1.prod.a.momentohq.com', null, {})
```
cprice404, line 15: is this just pseudocode or does it reflect what you actually expect the user to write?

The main reason that I ask is that we currently have this abstraction, `SimpleCacheClient`, that hides the generated `ScsClient`.  In the design review for SDK configuration, I asked if we ever expected end users to construct the lower-level ones directly, because it had some implications on the choice of where we would manage the gRPC channels in the cases where we are creating more than one.  I believe we decided that we would treat these as effectively private, and users would only construct the higher-level one.

cprice404 (commented):
cool.  just some questions on naming things etc.

javascript-web/README.md
kvcache: it's consistent with the other one; I think, such as it is, it's right.

javascript-web/README.md
kvcache: This is what I expect someone who uses the protos to write. We will use the protos, and others can too. It's okay if people want to use the protos - but we're not offering a whole lot of help to do so. "Effectively private" is fine - but they're open source and totally fair game =)

.github/workflows/pull_request.yml
kvcache: thanks - I missed some local changes that the ide didn't save -.-

cprice404 (approved):

PR #78, opened 2022-08-23
fix: take a single value as input for list push apis

anp13:
As per our discussion this morning, we want the push/pop operations to operate on a single value rather than multiple values. We already do that for the list pop apis. This PR updates the `ListPushFront` and `ListPushBack` api requests to take a single value as an input.


bruuuuuuuce (approved):
:shipit:

kvcache (approved):
this is an unsafe change from an encoding perspective. We don't have active usage with lists yet though so let's go ahead and do this now.

PR #80, opened 2022-08-25
feat: add truncate option to list push

### LINKED ISSUE #81: List push with truncate protos

malandis:
This adds an optional parameter to `ListPushFront` and `ListPushBack`
requests to maintain a list length invariant. Should the push
operation make the list exceed the given length, the excess is trimmed
from the opposite end of the list.

Closes #81

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -209,7 +209,7 @@ message _ListPushFrontRequest {
   bool refresh_ttl = 4;
 
   // ensure total length <= this; remove excess from back of list
-  optional uint64 truncate_tail_to_size = 5;
+  optional uint32 truncate_tail_to_size = 5;
 }
 
 message _ListPushFrontResponse {}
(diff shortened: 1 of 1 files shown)


bruuuuuuuce (approved):
:shipit:

proto/cacheclient.proto
```diff
@@ -207,6 +207,9 @@ message _ListPushFrontRequest {
   bytes value = 2;
   uint64 ttl_milliseconds = 3;
   bool refresh_ttl = 4;
+
+  // ensure total length <= this; remove excess from back of list
+  optional uint64 truncate_tail_to_size = 5;
```
anp13, line 212: we are using `uint32 ` for list length. Do you think we should update that?

malandis: Can do that. I was looking at the usage of `uint64` for `ttl_milliseconds`.

anp13: On second thoughts, I think `uint32` should suffice. The ttl is in milliseconds and that is why we used `uint64` there.

anp13 (approved):

PR #84, opened 2022-08-26
fix: only generate gRPC client code for C#

malandis:
The C# generator allows the option to generate client code vs server
code vs both. By default it generates both. Since the clients only
need the client code, this change narrows the build to only generate
those.


bruuuuuuuce (approved):
:shipit:

cprice404 (approved):

gautamomento (approved):

PR #85, opened 2022-08-29
fix: make truncate_*_to_size nomenclature consistent with ListPush*

malandis:
For `ListPush` We use the nouns "Front" and "Back" to refer to the
beginning and end of the list respectively. In the previous commit, in
the context of truncate, we used the nouns "Head" and "Tail". That
introduced extra vocabulary to refer to exactly the same
concepts. This commit renames "Head" and "Tail" to "Front" and "Back"
for consistency.


bruuuuuuuce (approved):
:shipit:

PR #86, opened 2022-08-30
feat: list push returns the length of the list post-push

malandis:
This adds the list length to `ListPushFrontResponse` and
`ListPushBackResponse`. This corresponds to the length of the list
just after the request's value is pushed.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -212,7 +212,10 @@ message _ListPushFrontRequest {
   optional uint32 truncate_tail_to_size = 5;
 }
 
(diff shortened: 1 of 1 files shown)


bruuuuuuuce (approved):
:shipit:

anp13 (commented):
Do you think it makes sense to have the list length in the list pop front/back api response as well?

bruuuuuuuce:
> Do you think it makes sense to have the list length in the list pop front/back api response as well?

Yea I like that idea

anp13 (approved):

PR #88, opened 2022-09-09
feat: add `DictionaryIncrement`

malandis:
This adds the RPC, request, and response messages for `DictionaryIncrement`. The response includes a `status` field to indicate whether the operation was successful or not.

An increment operation (`DictionaryIncrement` or regular `Increment`) may fail if the stored value does not parse to a number. Other errors to consider including are integer overflow or integer underflow.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -16,6 +16,10 @@ enum ECacheResult {
   reserved 4 to 6;
 }
 
+enum ECacheIncrementResult {
+  Ok = 1;
+  ParseError = 2;
+}
 
 service Scs {
   rpc Get (_GetRequest) returns (_GetResponse) {}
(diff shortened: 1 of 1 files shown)


cprice404 (approved):

bruuuuuuuce (approved):
:shipit:

proto/cacheclient.proto
```diff
@@ -122,6 +127,19 @@ message _DictionarySetRequest {
...
+message _DictionaryIncrementRequest {
+  bytes dictionary_name = 1;
+  bytes field = 2;
+  int64 amount = 3;
+  uint64 ttl_milliseconds = 4;
+  bool refresh_ttl = 5;
+}
+
+message _DictionaryIncrementResponse {
+  ECacheIncrementResult status = 1;
```
anp13, line 139: Do we need this status? I think the plan was to even deprecate the `ECacheResult` and have oneof `Hit` or `Miss` in the response. We don't return a status for any of the other apis for error cases.

anp13 (commented):
Looks good. Just one question.

proto/cacheclient.proto
malandis: We need a way to communicate if the increment failed and why, in particular for parsing errorrs, overflow errors, and underflow errors. Do we have a preferred way to do this?

anp13: We use status object only for valid non-error cases. For error cases we return appropriate grpc status codes with apt error messages.

PR #89, opened 2022-09-09
fix: remove increment status code

malandis:
We will propagate increment failures with exceptions. For incrementing a value that is not an integer, we will use gRPC status code `FAILED_PRECONDITION`.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -16,11 +16,6 @@ enum ECacheResult {
   reserved 4 to 6;
 }
 
-enum ECacheIncrementResult {
-  IncrementOk = 0;
-  IncrementParseError = 1;
-}
-
 service Scs {
   rpc Get (_GetRequest) returns (_GetResponse) {}
   rpc Set (_SetRequest) returns (_SetResponse) {}
@@ -136,8 +131,7 @@ message _DictionaryIncrementRequest {
 }
 
 message _DictionaryIncrementResponse {
-  ECacheIncrementResult status = 1;
-  int64 value = 2;
+  int64 value = 1;
 }
 
 message _DictionaryDeleteRequest {


bruuuuuuuce (approved):
:shipit:

cprice404:
just curious why you prefer grpc status code to the status object?

proto/cacheclient.proto
```diff
@@ -136,8 +131,7 @@ message _DictionaryIncrementRequest {
 }
 
 message _DictionaryIncrementResponse {
-  ECacheIncrementResult status = 1;
-  int64 value = 2;
+  int64 value = 1;
```
anp13, line 134: For `DictionaryIncrement`, the behavior would be the same as `DictionarySet` where we will create the dictionary if it does not exist and add the field to it and initialize the field to 0. Correct? 

Also should the `amount` provided in the request and the `value` returned in the response have type `uint` instead of `int`?

malandis: @cprice404 https://github.com/momentohq/client_protos/pull/88#pullrequestreview-1102826604

cprice404: re: uint - Michael's docs state that both can be negative, which seems reasonable to me.

malandis: @anp13 Yes same behavior for `DictionarySet`.

`uint` vs `int64`: while the name of the command is `increment`, the semantics are `+=`. Thus positive, negative, and zero amounts are all valid.

For more details see the spec on notion and [C# documentation](https://github.com/momentohq/client-sdk-csharp/blob/main/src/Momento.Sdk/Incubating/SimpleCacheClient.cs#L240-L277)

anp13: Ohh. So user can use this api to decrement the value as well? The name says "increment" but I guess technically you can use a negative integer to decrement the value as well. Got it.

anp13 (approved):

cprice404:
> > just curious why you prefer grpc status code to the status object?
> 
> We use status object only for valid non-error cases. For error cases we return appropriate grpc status codes with apt error messages.

Okay... I understand the rationale there I guess, and I'm in favor of consistency, but one thing that I will call out here is that there have been places in the SDKs where error handling is difficult because a given grpc status code can mean multiple things.  e.g., RESOURCE_EXHAUSTED and TIMEOUT.

In cases like those, it is difficult to give the user a helpful message that allows them to clearly understand what went wrong and how to fix it.

In this case, if there is no ambiguity because there is no other way that we can trigger a FAILED_PRECONDITION, then it's okay.  But I would ask two things going forward:

1. When the server throws one of these errors please try to make it include a helpful message that we can consider passing through to the user, and
2. If there are cases where the gRPC error type would be ambiguous we should re-evaluate our dogma here, and make the decision that is best for the end user.

malandis:
@cprice404 @anp13 Thanks for the insightful comments. At least for this one case we have an unambiguous error code. If this becomes difficult then we can add in status codes in the future.

PR #94, opened 2022-09-22
fix: update grpc versions to match what we have in kitchensink

bruuuuuuuce:
Our shared libraries are using newer versions of grpc than what our client_protos were using. This was causing some conflicting jars to be pulled, preventing us from being able to send metrics from our canaries. This pr addresses this by aligning the versions of grpc with what we have inside of kitchensink


cprice404 (approved):
thanks for the helpful commit message and thanks for tackling this!!

untangle

PR #95, opened 2022-11-10
feat: update cacheclient with new list APIs

tylerburdsall:
PR that captures incoming APIs: `list_concatenate_front` and `list_concatenate_back`

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -222,7 +222,7 @@ message _ListConcatenateFrontRequest {
   bool refresh_ttl = 4;
 
   // ensure total length <= this; remove excess from back of list
-  optional uint32 truncate_back_to_size = 5;
+  uint32 truncate_back_to_size = 5;
 }
 
 message _ListConcatenateFrontResponse {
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -213,6 +215,37 @@ message _SetDifferenceResponse {
...
 }
 
+message _ListConcatenateFrontRequest {
+  bytes list_name = 1;
+  repeated bytes values = 2;
+  uint64 ttl_milliseconds = 3;
+  bool refresh_ttl = 4;
+
+  // ensure total length <= this; remove excess from back of list
+  optional uint32 truncate_back_to_size = 5;
```
bruuuuuuuce, line 225: these no longer need to be optional? Since everything is by default optional in proto3

tylerburdsall: Oh good to know! I can adjust these then

tylerburdsall: addressed in https://github.com/momentohq/client-protos/pull/95/commits/28f45939251d30973cd38d3af2d30ea607c0cbb9

bruuuuuuuce (approved):
:shipit:

PR #96, opened 2022-12-01
chore: update grpc-js library to latest

cprice404:
We have encountered some minor issues with grpc-js running in
AWS Lambda environment.  We're a little out of date on our
grpc dependency so this brings us up to the latest.  This
didn't fix the AWS lambda issues, but does include some bug
fixes and updates that seem worth picking up.  See release notes
here:

https://github.com/grpc/grpc-node/releases

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -13,7 +13,6 @@
     "build": "rm -rf dist && mkdir dist && ./generate_protos.sh && tsc"
   },
   "author": "",
-  "license": "MIT",
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


javascript/package.json
```diff
@@ -13,17 +13,16 @@
     "build": "rm -rf dist && mkdir dist && ./generate_protos.sh && tsc"
   },
   "author": "",
-  "license": "MIT",
```
poppoerika, line 16: Was removing `license` property intentional?

poppoerika (approved):

PR #97, opened 2022-12-07
feat: Add proto for SetMultipleIfNotExists

rlinehan:
`SetMultipleIfNotExists` allows for setting multiple key-value pairs. Each will be set only if the key does not exist (was not previously set).

- Redis' behavior for [`MSETNX`](https://redis.io/commands/msetnx/) is to only set any keys if no keys exists. If no keys exist then all keys will be set, but if any key exists then no keys will be set. It's an atomic set of all keys.
- With a multi-node set up, we can't easily support atomic set for multiple keys.
- Instead, we will set any keys that don't exist.
- Since we're doing this behavior on each key, we'll return a result for each key - whether it was stored or not.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -21,6 +21,8 @@ service Scs {
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
 
+  rpc SetMultipleIfNotExists (_SetMultipleIfNotExistsRequest) returns (_SetMultipleIfNotExistsResponse) {}
+
   rpc DictionaryGet (_DictionaryGetRequest) returns (_DictionaryGetResponse) {}
   rpc DictionaryFetch (_DictionaryFetchRequest) returns (_DictionaryFetchResponse) {}
   rpc DictionarySet (_DictionarySetRequest) returns (_DictionarySetResponse) {}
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -70,6 +72,29 @@ message _SetResponse {
...
+  repeated _KeyValuePair items = 1;
+  uint64 ttl_milliseconds = 2;
+  bool refresh_ttl = 3;
+}
+
+message _SetMultipleIfNotExistsResponse {
+  message _SetMultipleIfNotExistsResponsePart {
+    oneof result {
+      Ok ok = 1;
+      NotStored not_stored = 2;
```
rlinehan, line 90: Hm... wondering if these should be `_Ok` and `_NotStored`, based on https://github.com/momentohq/client-protos/blob/04cd7201a2f8bbba3c2011a1f509ca7373538699/proto/cacheclient.proto#L115-L118

proto/cacheclient.proto
```diff
@@ -70,6 +72,29 @@ message _SetResponse {
...
+    }
+    message Ok { }
+    message NotStored { }
```
rlinehan, line 93: Hm... wondering if these should be `_Ok` and `_NotStored`, based on https://github.com/momentohq/client-protos/blob/04cd7201a2f8bbba3c2011a1f509ca7373538699/proto/cacheclient.proto#L115-L118

proto/cacheclient.proto
```diff
@@ -70,6 +72,29 @@ message _SetResponse {
   string message = 2;
 }
 
+message _KeyValuePair {
+  bytes field = 1;
+  bytes value = 2;
+}
```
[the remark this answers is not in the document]
malandis: Regarding the value of the names: for your addition we'd want the fields to be `key`, and `value`, whereas for the dictionary one we want `field` and `value`. The name `field` matters to have a word to distinguish the cache key from a dictionary key.

While the messages have the same content -- two fields of type `bytes` -- when programming against generated code, it's really nice to have specific names.

rlinehan:
Closing as we've decided not to do this quite yet.

PR #99, opened 2022-12-09
feat: Add an RPC to support flushing caches

gautamomento:
The RPC will allow callers to flush all the data from a given cache. Once this operation is invoked on a cache all the data prior to invoking the operation will not be available to the callers.

### DIFF
--- a/proto/controlclient.proto
+++ b/proto/controlclient.proto
@@ -11,6 +11,7 @@ service ScsControl {
   rpc CreateCache (_CreateCacheRequest) returns (_CreateCacheResponse) {}
   rpc DeleteCache (_DeleteCacheRequest) returns (_DeleteCacheResponse) {}
   rpc ListCaches (_ListCachesRequest) returns (_ListCachesResponse) {}
+  rpc FlushCache (_FlushCacheRequest) returns (_FlushCacheResponse) {}
   rpc CreateSigningKey (_CreateSigningKeyRequest) returns (_CreateSigningKeyResponse) {}
   rpc RevokeSigningKey (_RevokeSigningKeyRequest) returns (_RevokeSigningKeyResponse) {}
   rpc ListSigningKeys (_ListSigningKeysRequest) returns (_ListSigningKeysResponse) {}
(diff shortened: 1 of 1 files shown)


mjindalmotiv8 (approved):

proto/controlclient.proto
```diff
@@ -67,10 +68,18 @@ message _SigningKey {
...
+  string next_token = 1;
 }
 
 message _ListSigningKeysResponse {
   repeated _SigningKey signing_key = 1;
   string next_token = 2;
 }
+
+message _FlushCacheRequest {
+  string cache_name = 1;
```
cprice404, line 80: do we need cache name here?

cprice404: nm I was getting confused with data plane APIs where we are sending that as a header.

cprice404 (approved):

proto/controlclient.proto
```diff
@@ -67,10 +68,18 @@ message _SigningKey {
 }
 
 message _ListSigningKeysRequest {
-  string next_token = 1; 
+  string next_token = 1;
 }
```
kvcache, line 72: my kingdom for a `returns (stream _ListSigningKeysResponse)`

proto/controlclient.proto
```diff
@@ -11,6 +11,7 @@ service ScsControl {
   rpc CreateCache (_CreateCacheRequest) returns (_CreateCacheResponse) {}
   rpc DeleteCache (_DeleteCacheRequest) returns (_DeleteCacheResponse) {}
   rpc ListCaches (_ListCachesRequest) returns (_ListCachesResponse) {}
+  rpc FlushCache (_FlushCacheRequest) returns (_FlushCacheResponse) {}
```
kvcache, line 14:
```suggestion
  // Clear the contents of the cache. This is not transactional, but it should be
  // honored within seconds of the response.
  rpc FlushCache (_FlushCacheRequest) returns (_FlushCacheResponse) {}
```

kvcache (approved):

malandis (approved):

PR #101, opened 2022-12-13
feat: move ping RPC to separate service

cprice404:
The main cache service requires a "cache" header with a valid
cache name on every request.  This doesn't work well for the
Ping RPC because we won't necessarily know a cache name (and we
really don't need to know one) at the time we issue the request.

Rather than hacking a bunch of special case logic into the
already-fairly-complex auth interceptor for the cache service,
this moves the ping RPC to a separate service where we can
run without those requirements.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -16,9 +16,14 @@ enum ECacheResult {
   reserved 4 to 6;
 }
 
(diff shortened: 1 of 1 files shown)


kvcache (approved):
cool

PR #102, opened 2022-12-14
feat: move ping service to separate proto file

cprice404:
For some reason, the stubs generated for the server-side for kotlin were very messed up when there were two services defined in the same file, and it was impossible to bind a separate service to the ping implementation.  This commit moves the ping protos to a new file.

I've tested this via composite build locally and it is working.


kvcache (approved):
..cute. I wish I knew why it did that to the reflection service.

PR #103, opened 2022-12-14
feat: Add protos for SetIfNotExists API 

rlinehan:
Part of https://github.com/momentohq/cacheservice/issues/1092

~~`add`~~ `SetIfNotExists` will store the value for the key if it isn't yet present, and
return `stored`. If the key is already present it will not overwrite the
value and will return `not stored.`


proto/cacheclient.proto
```diff
@@ -19,6 +19,7 @@ enum ECacheResult {
 service Scs {
   rpc Get (_GetRequest) returns (_GetResponse) {}
   rpc Set (_SetRequest) returns (_SetResponse) {}
+  rpc Add (_AddRequest) returns (_AddResponse) {}
```
rlinehan, line 22: Alternate name for this could be `SetIfNotExists`, if people would like to discuss the name. I went with `Add` because it's 1) shorter, 2) seems to be a pretty common action in other caches so won't be unfamiliar to users.

kvcache (approved):
Looks good

proto/cacheclient.proto
cprice404: I feel like `SetIfNotExists` is more clear in its intention, but I don't feel that strongly about it, so I'm okay with deferring to majority of most folks prefer `Add`.

cprice404 (dismissed):

proto/cacheclient.proto
malandis: Agreed. Would have preferred http verbs from the beginning -- `put` for `set`; then this would be `post`. I know `Add` has history with memcache; but having `Add` and (eventually) `Increment` seems confusing. Don't feel very strongly about it though.

rlinehan: Thinking about this more: I was originally biased by the fact that in MR2 code we already call this `Add`. Our users won't have that bias. Although `Add` may be familiar to some users who have experience with other caches, `SetIfNotExists` will be easily understood by anyone because it is very as to what it does. `Add` requires people to read docs to be certain of what it does, `SetIfNotExists` does not.

kvcache: I am more opinionated that it doesn't matter very much and that either is reasonable. I'm just going to throw behind whichever is more likely to carry the day. If you want to change it okay but don't waffle back to Add

cprice404 (approved):

PR #104, opened 2023-01-11
feat: Add protos for Integer type

rlinehan:
Add protos for APIs for get, set, and increment on integer type.

This is called `Integer` rather than `Counter` because it can be assigned in addition to being incremented. It is `Integer` so that should we ever want to add support for floats, we can add `Number`.

Although this is not part of the protos, a few implementation details:
- If increment is called on an integer that does not yet exist, it will be initialized as 0 first.
- ~~If increment is called without a provided value to increment by, it will default to incrementing by 1.~~ Default will be 0 because that is what proto3 has for missing value for int64 - that is the service will increment by whatever value it receives. If we want we could have SDKs provide 1 as a default.
- The value to increment by can be negative (that is, `Increment` will do all the things the Redis `INCR`, `DECR`, `INCRBY`, and `DECRBY` commands do, in one API.)

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -42,6 +42,10 @@ service Scs {
   rpc ListLength(_ListLengthRequest) returns (_ListLengthResponse) {}
   rpc ListConcatenateFront(_ListConcatenateFrontRequest) returns (_ListConcatenateFrontResponse) {}
   rpc ListConcatenateBack(_ListConcatenateBackRequest) returns (_ListConcatenateBackResponse) {}
+
+  rpc IntegerGet(_IntegerGetRequest) returns (_IntegerGetRequest) {}
+  rpc IntegerIncrement(_IntegerIncrementRequest) returns (_IntegerIncrementResponse) {}
+  rpc IntegerSet(_IntegerSetRequest) returns (_IntegerSetResponse) {}
 }
 
 message _GetRequest {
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -391,3 +395,40 @@ message _ListLengthResponse {
...
+
+  oneof integer {
+    _Found found = 1;
+    _Missing missing = 2;
+  }
+}
+
+message _IntegerIncrementRequest {
+  bytes integer_name = 1;
+  int64 increment_by = 2;
```
kvcache, line 418: > If increment is called without a provided value to increment by, it will default to incrementing by 1.

This will cause a weirdness where if someone does not do a special-case check for `0` we will increment by 1. They might truly want to call increment(0) because of details of their software. Remember, "unset" in `proto3` syntax for the int64 type is the nonnull value `0`.

bruuuuuuuce (commented):
Just a couple thoughts on this. I am aligned that we should only have 1 api to do increment and decrement. Do we know what this will look like from an sdk perspective?

```
momento.increment("my-key")
momento.intIncrement("my-key")
```

Also, as mentioned, this seems to open the door for us to add a `momento.numberIncrement` method. Do we think it would be confusing to users to have both a `numberIncrement` and an `intIncrement` method?

proto/cacheclient.proto
pete-naylor: Maybe it is safest to require a value for increment_by ?

kvcache: It is not possible to not provide a value for increment_by in `proto3`. A missing value on the serialized bits sent over the network is equivalent to setting `0`. https://developers.google.com/protocol-buffers/docs/proto3#default and some implementations may in fact elide the bits for this field when `0` was explicitly set.

You could use `oneof` and define messages like `message Increment {}` `message Add { int64 amount = 1; }` to explicitly choose the intended semantic

rlinehan: > Remember, "unset" in `proto3` syntax for the int64 type is the nonnull value `0`.

TIL. I did not know this. Agreed that this means having default if no value is provided be 1 does not make sense. We could have that be the case in the SDKs if we want, but agreed it should not be the case from server's perspective.

kvcache: I am supportive of letting the number be what it is and doing "add" with the provided number on the service. I agree that the SDK can form the message however it likes.

cprice404:
We are representing integers as strings for the purposes of the dictionary increment operation.  I realize that this was not a unanimous or well-loved decision.  But do we not have concerns about the inconsistency of using actual integers here vs. what we are doing there?

cprice404 (commented):
i don't feel too strongly but wanted to raise the question about consistency.  i had imagined that this would be an almost direct copy/paste of the dictionary increment.

rlinehan:
After further discussion and some playing around with the dotnet SDK, I'm closing this in favor of #106, which just adds `Increment`.

We don't have "types" - we have datastructures. We have no actual restrictions on storing e.g. a key/value pair where the key is the name of a dictionary... we sometimes throw errors, but not consistently (and currently these are actually `Unavailable` rather than saying anything about bad type). Adding an `Integer` type where it would actually fully be a type (e.g. got errors if you tried to call `get` on it rather than `getInteger`) would be hard to do and would also not be consistent with how we've done this for `DictionaryIncrement` on a field.

PR #105, opened 2023-01-13
feat: add pubsub proto definitions

kvcache:
For preview and development, here is a possible definition for a pubsub feature with Momento caches.

A topic lives within a Momento cache. Using the Pubsub service, users can subscribe and publish to topics. Topics are multicast message relays. They are lossy, but they try to let you know when you might have missed something.

### DIFF
--- a/proto/cachepubsub.proto
+++ b/proto/cachepubsub.proto
@@ -1,0 +1,126 @@
+syntax = "proto3";
+
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_multiple_files = true;
+option java_package = "grpc.cache_client.pubsub";
+option csharp_namespace = "Momento.Protos.CacheClient.Pubsub";
+
+package cache_client.pubsub;
+
+// For working with topics in a cache.
+// Momento topics are conceptually located on a cache. They are best-effort multicast.
+// To use them, create a cache then start subscribing and publishing!
+//
+// Momento topic subscriptions try to give you information about the quality of the
+//   stream you are receiving. For example, you might miss messages if your network
+//   is slow, or if some intermediate switch fails, or due to rate limiting. It is
+//   also possible, though we try to avoid it, that messages could briefly come out
+//   of order between subscribers.
+//   We try to tell you when things like this happen via a Discontinuity in your
+//   subscription stream. If you do not care about occasional discontinuities then
+//   don't bother handling them! You might still want to log them just in case ;-)
+service Pubsub {
+  // Publish a message to a topic.
+  //
+  // If a topic has no subscribers, then the effect of Publish MAY be either of:
+  // * It is dropped and the topic is nonexistent.
+  // * It is accepted to the topic as the next message.
+  //
+  // Publish() does not wait for subscribers to accept. It returns Ok upon accepting
+  // the topic value. It also returns Ok if there are no subscribers and the value
+  // happens to be dropped. Publish() can not guarantee delivery in theory but in
+  // practice it should almost always deliver to subscribers.
+  //
+  // REQUIRES HEADER authorization: Momento auth token
+  rpc Publish(_PublishRequest) returns (_Empty);
+
+  // Subscribe to notifications from a topic.
+  //
+  // You will receive a stream of values and (hopefully occasional) discontinuities.
+  // Values will appear as copies of the payloads you Publish() to the topic.
+  //
+  // REQUIRES HEADER authorization: Momento auth token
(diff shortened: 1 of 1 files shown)


proto/cachepubsub.proto
```diff
@@ -0,0 +1,126 @@
...
+  string cache_name = 1;
+
+  // The literal topic name to which you want to subscribe.
+  string topic = 2;
+
+  // --> Providing this is not required. <--
+  //
+  // If provided, attempt to reconnect to the topic and replay messages starting from
+  // the provided sequence number. You may get a discontinuity if some (or all) of the
+  // messages are not available.
```
cprice404, line 71: based on what you told me about how you plan to manage the messages in memory, this seems like it would rarely ever return many missed messages.  What am I missing?

(In theory I like this as part of the API design but it just doesn't seem like it would provide much value in most cases?)

proto/cachepubsub.proto
```diff
@@ -0,0 +1,126 @@
...
+message _SubscriptionItem {
+  oneof kind {
+    // The subscription has yielded an item you previously published.
+    _TopicItem item = 1;
+    // Momento wants to let you know we detected some possible inconsistency at this
+    // point in the subscription stream.
+    //
+    // A lack of a discontinuity does not mean the subscription is guaranteed to be
+    // strictly perfect, but the presence of a discontinuity is very likely to
+    _Discontinuity discontinuity = 2;
```
cprice404, line 88: do we need to support the idea of a topic being "closed" by the publisher?  i presume that would go here

cprice404 (commented):
looks great just a couple of small questions

proto/cachepubsub.proto
virratanasangpunth, line 88: do we proactively send down a discontinuity whenever we detect one? (probably mr2 who will detect these is my guess).



or do we do this only when the client provides us with `resume_at_topic_sequence_number`

virratanasangpunth (approved):

proto/cachepubsub.proto
kvcache: Our initial implementation will not have an in-memory topic buffer. Any deployment to the front-end service will cause a reconnect, for which you will almost surely receive a discontinuity. You do not need to do anything about it today, but I'd recommend at least logging so we can more easily troubleshoot issues (e.g., is your subscription simply not working or are you getting a ton of gaps in your stream?)

Similarly, a backend deployment will cause a relocation of the Topic's source of truth. The coordination of a happy-path quiesce point and resume is not going to fit in our early milestones either.

Discontinuities are a good way for users to consider whether momento topics are giving them a good enough quality of subscription relative to other potential lossy pubsub services (or applications), and a good way for us to measure when we're doing well and when we could improve.

proto/cachepubsub.proto
kvcache: I don't have a "close" verb. The current way we have modeled to close a topic is to quit publishing to it and quit subscribing to it. I'd be open to supporting `close` but do you think we should wait for it to have demonstrable value first?

kvcache: our backend can detect certain kinds of discontinuities while our frontend would know about others. Either can relay a discontinuity. Wherever you'd `log { msg("we lost some data") }` you can `emit(discontinuity(fromNum, toNum))` instead and give the stream more rich metadata about its quality. It'll be nice for monitoring

proto/cachepubsub.proto
kvcache: To be fair, deployments of either service role should be pretty rare relative to a popular Topic's event rate; and when nothing is broken or relocating there should be no discontinuities.

proto/cachepubsub.proto
cprice404: that's fine, just making sure you had thought about where it would fit in.

proto/cachepubsub.proto
cprice404: i haven't spent enough time thinking about deployments.  we should talk at some point about what the implications of this are in terms of our ability to deploy mr2 and ms2 and if there are any design decisions we can make that will help us maintain a relatively easy story for CD.

cprice404 (approved):

proto/cachepubsub.proto
kvcache: Gotcha - yes, this is exactly where other kinds of control messages or metadata messages would go.
TopicValue is where we'll add any future service-honored message representations (could make some pretty neat add-ons there).

PR #106, opened 2023-01-13
feat: Add Increment API

rlinehan:
Add proto definitions for rpc, request, and response for `Increment`.

`Increment` takes a key, an integer by which to increment by (can be negative to decrement), and a ttl. If the key does not already exist, it initializes it at 0.

The `Increment` operation maybe may fail if the stored value (which is stored as a string) does not parse to a number. Currently this error will surface as an `Unavailable` error; in my opinion in order for this API to be usable we will want to make sure we surface the full error.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -139,6 +139,16 @@ message _DictionarySetRequest {
 
 message _DictionarySetResponse {}
 
+message _IncrementRequest {
+  bytes cache_key = 1;
+  int64 increment_by = 2;
+  uint64 ttl_milliseconds = 3;
+}
+
+message _IncrementResponse {
+  int64 value = 1;
+}
+
 message _DictionaryIncrementRequest {
   bytes dictionary_name = 1;
   bytes field = 2;
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -141,11 +141,14 @@ message _DictionarySetResponse {}
 
 message _IncrementRequest {
   bytes cache_key = 1;
+  // Amount to add to the stored value.
+  // If this key doesn't currently exist, it's created with this value (encoded as a base 10 string)
   int64 increment_by = 2;
   uint64 ttl_milliseconds = 3;
 }
 
 message _IncrementResponse {
+  // The value stored after the increment operation.
   int64 value = 1;
 }
 
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -21,6 +21,7 @@ service Scs {
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {}
   rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
+  rpc Increment (_IncrementRequest) returns (_IncrementResponse) {}
 
   rpc DictionaryGet (_DictionaryGetRequest) returns (_DictionaryGetResponse) {}
   rpc DictionaryFetch (_DictionaryFetchRequest) returns (_DictionaryFetchResponse) {}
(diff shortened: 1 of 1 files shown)


bruuuuuuuce:
> Currently this error will surface as an Unavailable error; in my opinion in order for this API to be usable we will want to make sure we surface the full error.

For dictionary increment, we return the grpc error code `FAILED_PRECONDITION` when this happens

proto/cacheclient.proto
```diff
@@ -139,6 +139,16 @@ message _DictionarySetRequest {
 
 message _DictionarySetResponse {}
 
+message _IncrementRequest {
+  bytes cache_key = 1;
+  int64 increment_by = 2;
+  uint64 ttl_milliseconds = 3;
```
bruuuuuuuce, line 145: originally I was going to suggest here to add a `bool refresh_ttl = 4;`, but thinking about it more, I dont think it makes sense for this. Because inside of our backend, this is essentially a special unary operation, and for unary sets, we also don't have a refresh ttl semantic. So I think this looks good 👍

bruuuuuuuce (approved):
:shipit:

proto/cacheclient.proto
```diff
@@ -139,6 +139,16 @@ message _DictionarySetRequest {
...
+}
+
+message _IncrementResponse {
+  int64 value = 1;
+}
```
kvcache, line 150:
```suggestion
message _IncrementResponse {
  // The value stored after the increment operation.
  int64 value = 1;
}
```

proto/cacheclient.proto
```diff
@@ -139,6 +139,16 @@ message _DictionarySetRequest {
 
 message _DictionarySetResponse {}
 
+message _IncrementRequest {
+  bytes cache_key = 1;
+  int64 increment_by = 2;
```
kvcache, line 144:
```suggestion
message _IncrementRequest {
  bytes cache_key = 1;
  // Amount to add to the stored value.
  // If this key doesn't currently exist, it's created with this value (encoded as a base 10 string)
  int64 increment_by = 2;
```

I really like proto comments. They even make it into generated code in all languages I've used.

proto/cacheclient.proto
```diff
@@ -139,6 +139,16 @@ message _DictionarySetRequest {
 
 message _DictionarySetResponse {}
 
+message _IncrementRequest {
```
kvcache, line 142: I don't see the `rpc` definition in Scs above. Is that intended to come later?

nit: These types are sandwiched between some dictionary types. Did you consider putting them maybe up by the other scalar definitions or down at the bottom instead?

kvcache (commented):
I know we have been really lax about this up to now but it's probably worth commenting protos with semantics.

proto/cacheclient.proto
rlinehan: 🤦 yep, both of those concerns were intended to be fixed before I pushed.

proto/cacheclient.proto
rlinehan: Yeah, I had thought about it too and also settled on not having it since this is a basically "unary" operation and `refresh_ttl` is for when you have an action that handles part of a datastructure where the overall datastructure has a ttl (more or less). I'll add a comment to the commit message though about this.

proto/cacheclient.proto
rlinehan: Nice! Definitely agreed - haven't seen them before in our protos but they seem like a really nice thing to have. Thanks for these!

proto/cacheclient.proto
rlinehan: Fixed in [fff7562](https://github.com/momentohq/client-protos/pull/106/commits/fff7562676df26d8f28a1a9443b1289beb469f85), thanks for catching those two issues!

bruuuuuuuce (approved):
:shipit:

PR #113, opened 2023-01-26
fix: remove unknown as valid order option, fix typos

bruuuuuuuce:
renames `elemnt` to `element` for sorted set remove elements request

removes `unknown` as a valid option for when fetching a sorted set

https://gomomento.slack.com/archives/C03HUJ5GPAB/p1674774391650909

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -514,7 +514,7 @@ message _SortedSetRemoveRequest {
   message _Some {
     repeated bytes element_name = 1;
   }
-  oneof remove_elemnts {
+  oneof remove_elements {
     _All all = 2;
     _Some some = 3;
   }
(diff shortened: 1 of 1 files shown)


rlinehan (approved):

proto/cacheclient.proto
```diff
@@ -451,9 +451,8 @@ message _SortedSetPutResponse {}
 
 message _SortedSetFetchRequest {
   enum Order {
-    UNKNOWN = 0;
-    ASCENDING = 1;
-    DESCENDING = 2;
+    ASCENDING = 0;
+    DESCENDING = 1;
   }
```
kvcache, line 456: This renumbering will break any existing code's compatibility with the server. If there's nothing using this yet that's fine, but if there's any client using this it needs to be accounted for - is this currently unused?

bruuuuuuuce: Correct this is currently unused, this is only for the new sorted sets api which is unimplemented. In protobuf, enums need to start with 0, so if we want to remove `UNKNOWN` we need to renumber the enums

kvcache (approved):

PR #114, opened 2023-02-06
chore: Rename `increment_by` to `amount` for `Increment` API

rlinehan:
`DictionaryIncrement` and `SortedSetIncrement` use `amount` as the parameter for the value by which to increment, by scalar `Increment` currently uses `increment_by`. This updates scalar `Increment` to use `amount` as well for consistency, rather than `increment_by`.


bruuuuuuuce (approved):
:shipit:

cprice404 (approved):

PR #118, opened 2023-02-14
feat: add sorted sets fetch by score proto

wandaitzuchen:


proto/cacheclient.proto
```diff
@@ -455,18 +455,45 @@ message _SortedSetFetchRequest {
...
 
+  // num_results fields
   message _All {}
 
   message _Limit {
     uint32 limit = 1;
   }
 
+  // range fields
+  message _ByIndex {}
```
gautamomento, line 466: Should this also have structures similar to endpoints?

proto/cacheclient.proto
```diff
@@ -455,18 +455,45 @@ message _SortedSetFetchRequest {
...
+
+  message _ByScore {
+    message _Endpoint {
+      message _Score {
+        double score = 1;
+        bool open = 2;
```
gautamomento, line 472: A comment would be useful. I believe this means inclusive of the score or not?

gautamomento (dismissed):
Just some nits but LGTM.

proto/cacheclient.proto
wandaitzuchen: What's already implemented is "ByIndex", and it doesn't require additional arguments.

proto/cacheclient.proto
wandaitzuchen: Added! thanks.

gautamomento (approved):

PR #119, opened 2023-02-14
chore: add heartbeats to topic subscriptions

kvcache:
A tool for helping to make long-lived subscriptions to low-rate
topics.

### DIFF
--- a/proto/cachepubsub.proto
+++ b/proto/cachepubsub.proto
@@ -80,12 +80,16 @@ message _SubscriptionItem {
   oneof kind {
     // The subscription has yielded an item you previously published.
     _TopicItem item = 1;
+
     // Momento wants to let you know we detected some possible inconsistency at this
     // point in the subscription stream.
     //
     // A lack of a discontinuity does not mean the subscription is guaranteed to be
     // strictly perfect, but the presence of a discontinuity is very likely to
     _Discontinuity discontinuity = 2;
+
+    // The stream is still working, there's nothing to see here.
+    _Heartbeat heartbeat = 3;
   }
 }
 
@@ -123,4 +127,14 @@ message _Discontinuity {
   uint64 last_topic_sequence = 1;
   // The new topic sequence number after which TopicItems will ostensibly resume.
   uint64 new_topic_sequence = 2;
+}
+
+// A message from Momento for when we want to reassure clients or frameworks that a
+// Subscription is still healthy.
+// These are synthetic meta-events and do not increase the topic sequence count.
+// Different subscribers may receive a different cadence of heartbeat, and no guarantee
+// is made about the cadence or even presence or absence of heartbeats in a stream.
+// They are a tool for helping ensure that socket timeouts and the like don't impact
+// subscriptions you may care about, but that aren't receiving a substantial publish rate.
+message _Heartbeat {
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.idea
+.vscode


eaddingtonwhite (approved):
LGTM. Looks like pr failing since from fork I think.

tylerburdsall (approved):

proto/cachepubsub.proto
```diff
@@ -124,3 +128,13 @@ message _Discontinuity {
   // The new topic sequence number after which TopicItems will ostensibly resume.
   uint64 new_topic_sequence = 2;
 }
+
+// A message from Momento for when we want to reassure clients or frameworks that a
+// Subscription is still healthy.
+// These are synthetic meta-events and do not increase the topic sequence count.
+// Different subscribers may receive a different cadence of heartbeat, and no guarantee
+// is made about the cadence or even presence or absence of heartbeats in a stream.
```
cprice404, line 136: just a question about the "guarantee" part here...  maybe a few questions.

1. In my head this seems like a message that we would handle entirely internally to the SDK, so an end user would never see this... right?
2. If so, is it accurate that we don't want to (or need to) make some guarantees about the frequency?  I would think we would want to code the SDKs to some consistent language-independent spec for how they should expect these to occur, which comes fairly close to defining a "guarantee", I think?

kvcache: > In my head this seems like a message that we would handle entirely internally to the SDK, so an end user would never see this... right?

This repository is OSS and anyone can consume our proto artifacts. People who use our SDKs definitely should be insulated from this message kind, yes, but this message is certainly accessible to our users!

> is it accurate that we don't want to (or need to) make some guarantees about the frequency?

Yes, that is what I am proposing. Do you need some heartbeat frequency guarantee? I think it's premature to spec _when_ they will occur. We know we need them in some cases, and we may elide them in others. We should not build a distributed dependency on an exact frequency, I don't think... do you?

> I would think we would want to code the SDKs to some consistent language-independent spec for how they should expect these to occur, which comes fairly close to defining a "guarantee", I think?

These messages do not presuppose how SDKs are implemented and do not imply anything for user-facing SDK S&P semantics. If you want to surface them, you certainly can but I'd hope we would handle these messages internally within each SDK.

"How to think about these messages" is what these comments are about. They are a tool for [1] SDK developers and [2] users who directly consume proto artifacts; and also for [3] users of our SDKs who dive in and want to read how the protocol is implemented.

> how they should expect these to occur

I'm trying to write in here "SDK's should expect that these can occur at any time, but they are not required to occur at any time." That sounds specifically like not-a-guarantee to me 😄 but if you like that word we can try to wordsmith an alternative?

Bear in mind when considering a "guarantee" semantic that the server cannot guarantee that every subscription will receive a message (errors happen), let alone a heartbeat, and that there is no need I've heard of to send heartbeats to "active" subscriptions.

The language-agnostic semantic I'd propose for SDK's is "you should log.debug/swallow these until you have a reason to do something with them."

cprice404: > If you want to surface them, you certainly can but I'd hope we would handle these messages internally within each SDK.

yes, that is what I was saying in my comment.  I do not intend to surface them.

> The language-agnostic semantic I'd propose for SDK's is "you should log.debug/swallow these until you have a reason to do something with them."

i see, so you are not imagining that the SDKs would need to be coded to know that "if I don't receive a heartbeat within 60s then something bad has happened", you are imagining that the presence of the heartbeats will simply prevent the stream reset error from occurring, but the SDKs don't need to do anything special with or have any special knowledge about the heartbeats.

If that's the case then I'm good with your current wording.

cprice404: aight 🚢

cprice404 (approved):

PR #120, opened 2023-02-15
feat: Add LTRIM/LRANGE api definitions

ben-kugler:
Adding list trim and range protobuf definitions, used https://www.notion.so/momentohq/List-2219670735f24aa38fdc3dd877d2d928 as template


proto/cacheclient.proto
```diff
@@ -435,6 +437,47 @@ message _ListLengthResponse {
   }
 }
 
+message _ListTrimRequest {
+  bytes list_name = 1;
+  optional sint32 start = 2;
+  optional sint32 end = 3;
+  optional sint32 step = 4;
```
PDXKimani, line 444: What do you envision step doing in this API? I believe we are looking for an API which drops elements before start and after end, leaving any elements in between - I'm not sure how a step factors into this?

proto/cacheclient.proto
```diff
@@ -435,6 +437,47 @@ message _ListLengthResponse {
...
+  message _Found {
+    repeated bytes values = 1;
+  }
+
+  message _Missing {}
+
+  oneof list {
+    _Found found = 1;
+    _Missing missing = 2;
+  }
```
PDXKimani, line 457: I believe this should just a be a simple empty response? Receipt of the response would indicate a successful trim, otherwise it would return an error?

proto/cacheclient.proto
```diff
@@ -435,6 +437,47 @@ message _ListLengthResponse {
...
+}
+
+message _ListRangeRequest {
+  bytes list_name = 1;
+  optional sint32 start = 2;
+  optional sint32 end = 3;
+  optional sint32 step = 4;
```
PDXKimani: I think in the barest definition of this API we need simply return the range between start and end - what is step intended to do here? Do we foresee a use case for this?

cprice404:
hi @ben-kugler !  thanks for doing this!

Before we merge this, would you mind taking a minute to add these to the SDK API methods spec on Notion?  We're trying to do a better job of starting the convo about what the SDK implementations will look like before we define the protos, in hopes of keeping them more aligned.

cprice404:
Please feel free to reach out in chat if you need any help/advice on getting started adding them to the notion doc.  And lmk when you've added them, we can ping a few people to weigh in and get aligned.

proto/cacheclient.proto
cprice404: I agree with this, was going to ask the same.

cprice404 (commented):
waiting for changes based on outcome of API discussion meeting from yesterday afternoon; please re-request me as a reviewer whenever you're ready!

proto/cacheclient.proto
```diff
@@ -435,6 +438,16 @@ message _ListLengthResponse {
...
 
+message _ListRetainRequest {
+  bytes list_name = 1;
+  optional uint32 count = 2;
+  optional sint32 start_index = 3;
+  optional uint64 ttl_milliseconds = 4;
+  optional bool refresh_ttl = 5;
+}
+
+message _ListRetainResponse {}
```
ben-kugler, line 449: Removed the response, I don't think we really talked about it, and if we wanted to return the trimmed result.

proto/cacheclient.proto
```diff
@@ -403,6 +404,8 @@ message _ListRemoveResponse {}
 
 message _ListFetchRequest {
   bytes list_name = 1;
+  optional uint32 count = 2;
+  optional sint32 start_index = 3;
```
ben-kugler, line 408: Summary from the conversation,
1. We can add a `end_index` in the future, then enforce either count or end_index. This would _probably_ be enforced based on the sdk language as some use count and others use end_index
2. `Step`, a signed int, can be added later if customers ask for the ability to reverse through the list and/or change the stride to get every nth element

cprice404: i'm not sure if we need to explicitly mark these as optional in the proto; I think for other "optional" fields we have been just relying on recognizing when they have the protobuf "default" value.  Can you check me on this, @kvcache ?

kvcache: According to the 1st line in this file, this is `proto3` syntax.

Absent field value semantics are defined in the `proto3` language guide https://protobuf.dev/programming-guides/proto3/#default 

Some observations:
* `optional` does not mean anything material in proto3.
* Some code generators do seem to use it.
* Some protoc versions do not accept `optional` in proto3 (some linting thing) and it has blocked our pipelines before (internal pipelines iirc) when people marked fields `optional`.
* None of our ~60 request/response message definitions in this file use `optional`.

Do you have a concrete reason to use `optional` in these fields?

ben-kugler: https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0
Optional is on by default, so not necessary, removing. Looks like I was looking at proto2 docs as it's the second link for `proto optional`

proto/cacheclient.proto
```diff
@@ -403,6 +404,8 @@ message _ListRemoveResponse {}
 
 message _ListFetchRequest {
   bytes list_name = 1;
+  sint32 start_index = 3;
+  uint32 count = 2;
```
kvcache, line 408:
```suggestion
  sint32 start_index = 2;
  uint32 count = 3;
```

kvcache: also
```suggestion
  // Inclusive. 0 (start of list) by default
  sint32 start_index = 2;
  // If unset or 0, this effectively means list.length()
  uint32 count = 3;
```

kvcache (commented):
looks fine, 1 quality nit

cprice404:
thanks @ben-kugler , let's also get @schwern or @malandis to approve this before it goes in.  I'm good when Kenny + one of those two gentlemen is.

proto/cacheclient.proto
```diff
@@ -403,6 +404,10 @@ message _ListRemoveResponse {}
 
 message _ListFetchRequest {
   bytes list_name = 1;
+  // Inclusive. 0 (start of list) by default
+  sint32 start_index = 2;
+  // If unset or 0, this effectively means list.length()
+  uint32 count = 3;
```
schwern, line 410: Two more things that might be worth noting here.

1. A negative index counts from the end of the list.
2. If count is > the number of elements to return, return as much as you can.

schwern (approved):

schwern (approved):

malandis:
Per our discussion, we need to allow for count to be unspecified to account for "up to and including the list boundary" (whether that be the end or start). In other places we have been using a "oneof" for this, where one of the options is a count and the other is a message that means not specified.

PDXKimani (approved):

cprice404 (approved):

PR #121, opened 2023-02-16
chore: support fetch by index with offset

wandaitzuchen:
https://github.com/momentohq/cacheservice/issues/1230

See slack discussion: https://gomomento.slack.com/archives/C01AC7JT370/p1676508581644679?thread_ts=1676498195.969539&cid=C01AC7JT370

Note that the existing "limit" field means the number of elements to fetch.  This PR adds an "offset" field to represent the start index.  To achieve something like "I want to fetch 70,001th element ~ 80,000th element", (limit, offset) would be (10,000, 70,000).  In client SDKs we can have customer use the API this way, or we can implement some API for "give me from 70,001th to 80,000th element" and have the SDK do the conversion from (70,001, 80,000) to (10,000, 70,000) when sending over proto messages.


bruuuuuuuce:
I am trying to map this change to the redis api we will be using internally. I think it would be `zrange key start stop BYSCORE LIMIT offset count`. If the user wants to get the 100 elements around a specific element, they would first need to do `getRank` to get the rank of the element, and they could use that rank to determine the offset? I guess i'm confused how they would get the offset without making additional api calls

proto/cacheclient.proto
```diff
@@ -459,7 +459,8 @@ message _SortedSetFetchRequest {
   message _All {}
 
   message _Limit {
-    uint32 limit = 1;
+    uint32 limit = 1;   // aka count; number of elements to return
```
gautamomento, line 462: Is there a reason why we are doing a range in here instead of the 

```
  oneof range {
    _ByIndex by_index = 5;
    _ByScore by_score = 6;
  }
```
It feels more natural to query the range.

wandaitzuchen: Sorted sets doing query by index by default, and hence what "limit" does is essentially same as the range endpoints. 
In redis, you can do
- `ZRANGE key start stop`, e.g. `ZRANGE myzset 5 100` gives you index between 5 and 100 inclusive
- `ZRANGEBYSCORE key start stop LIMIT offset count`, e.g.  `ZRANGEBYSCORE myzset -inf 1000 LIMIT 5 96`, gives you all elements from the smallest up to score == 1000, and only returns from between 5 and 100 among the range of elements.  
`ZRANGE` command does not have "limit" option since it can be achieved by the start and stop indexes, but for `ZRANGEBYSCORE`, "limit" works on top of start and stop score endpoints. 
In other words the limit in the proto indeed determine the number of results returned.  Internally in ms2, this limit thing is mapped to the start and stop for ZRANGE, and mapped to count and offset for ZRANGEBYSCORE.  

Only for fetching by score we had this additional option to pick the start and stop scores.  For the already implemented fetching by index, there's no additional options that one can pick.

wandaitzuchen: Not sure if I answered your question.. IMO `limit` is something universal, so it's fine to stay outside of the "range" field. For fetch-by-index, there's no other arguments one can provide, but for other fetch-by types, there can be additional arguments (for example, when fetch-by-lex, one need to give min and max item values for the range).

gautamomento: Just summarizing what you expressed, limit allows to restrict and offset the result set. When ByIndex is specified this allows us to specify offset and limits. This option can also apply to restrict the number of results when we get a range by score.

gautamomento (approved):

PR #122, opened 2023-02-16
feat: Add ordering for Sorted Set Rank operation

gautamomento:
Adds an optional Order enum to the SortedSetRank Request. Given an element whose rank needs to be determined, the new ordering parameter determines how the sorted set will be ordered when looking for an element's presence and its rank.

When Ascending order is specified the rank of the element will be its index when the sorted set scores are orderd from low to high. For descending order, the rank of the element is its index when the sorted set scores are ordered from high to low.


bruuuuuuuce (approved):
:shipit:

PR #123, opened 2023-02-16
chore: Update the names of Sorted Set messages

gautamomento:
A sorted set is a collection of elements and each element has a value and its associated score.


bruuuuuuuce (approved):
:shipit:

PR #124, opened 2023-02-21
feat: add protos for generating secure tokens

bruuuuuuuce:
This pr adds a new api to the control plane to generate an api token/refresh token for users to recieve tokens that have an expiration date.

signing up via the cli will look something like

```
momento token generate --endpoint <endpoint> --expiration 30d
```

it will return
```
{
  "api_token": <JWT>,
  "refresh_token": <refresh token>,
  "endpoint": "cell-4-us-west-2.a.prod.momentohq.com",
  "valid_until": <epoch seconds when token expires> 
}
```

### DIFF
--- a/proto/controlclient.proto
+++ b/proto/controlclient.proto
@@ -15,6 +15,7 @@ service ScsControl {
   rpc CreateSigningKey (_CreateSigningKeyRequest) returns (_CreateSigningKeyResponse) {}
   rpc RevokeSigningKey (_RevokeSigningKeyRequest) returns (_RevokeSigningKeyResponse) {}
   rpc ListSigningKeys (_ListSigningKeysRequest) returns (_ListSigningKeysResponse) {}
+  rpc GenerateToken(_GenerateTokenRequest) returns (_GenerateTokenResponse) {}
 }
 
 message _DeleteCacheRequest {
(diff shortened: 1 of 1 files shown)


proto/controlclient.proto
```diff
@@ -83,3 +84,16 @@ message _FlushCacheRequest {
 message _FlushCacheResponse {
 
 }
+
+message _GenerateTokenRequest {
+  string session_token = 1;
```
bruuuuuuuce, line 89: should we keep the session token here for now, or add it in later?

rlinehan: It's passed as part of the authentication headers, so it doesn't need to be part of the protos. We'll need to add some special handling in the momento auth context to be able to say "only session tokens can do this action" (our first authorization checks!) but that will be handled within control plane.

proto/controlclient.proto
```diff
@@ -83,3 +85,22 @@ message _FlushCacheRequest {
...
+  // how many seconds do you want the api token to be valid for?
+  uint32 valid_for_seconds = 2;
+  string subject = 3;
```
rlinehan, line 93: At least for now, I don't think we should have the `subject` be passed as a part of the request - instead the API will read it out of the token used to make the request. That is, you can't create an API token for someone else, only for yourself.

bruuuuuuuce: >  instead the API will read it out of the token used to make the request

the `token` you are referring to here is the session token?

rlinehan: yes, it will be a session token (although for first pass can be from any type of token we already accept, and we'll add the session-only restriction later)

rlinehan (approved):
🎉 :shipit:

kvcache (approved):

PR #125, opened 2023-02-22
chore: use min/max for fetch by score range endpoints

wandaitzuchen:
https://github.com/momentohq/cacheservice/issues/1230

For dev eco folks - this PR and a few previous PRs add new fields to the `SortedSetFetchRequest` message in order to support fetching sorted set elements by score and returning results from specific start index. 

What's already implemented in go sdk: https://github.com/momentohq/client-protos/blob/db9f8e0534eb1916e07e5ea41d3f1e03013d161e/proto/cacheclient.proto#L452-L470
What's being proposed (everything since the version that go sdk supports):  
https://github.com/momentohq/client-protos/blob/1a9db5d164542dc99d3cc89e5b86ff3324b35bf6/proto/cacheclient.proto#L469-L515

I'm sorry that a few PRs got merged before I checked with you folks - please let me know if the shape of the updated API looks alright to you, or let me know your questions/suggestions!


cprice404:
thanks for the ping, @wandaitzuchen !  @malandis can you take a quick peek at this?

wandaitzuchen:
Synced with @malandis offline - to support a negative stop index, we decide to deprecate `Limit` message, and have each By* message handles relevant fields. 

This PR also update the fetch response to be able to include only the value.

wandaitzuchen:
`_SortedSetFetchRequest`/`_SortedSetFetchResponse` proto messages are updated and they're not backward compatible with the go sdk, considering we don't have customers actively using this API.  @eaddingtonwhite let me know if you'd prefer that we make this change later.

eaddingtonwhite:
I think lets go for it. Would rather move quickly to target state. We still have SDK labeled as alpha so is ok in my head. @wandaitzuchen lets just coordinate on SDK and example/demo updates and try swarm on it quickly.

malandis:
Here is the model for fetching by index: https://github.com/momentohq/client-protos/pull/126. It's not merged yet but this should be the final form (or very close to it)

cprice404:
@malandis @eaddingtonwhite are we good with this now?

eaddingtonwhite (dismissed):
I have coordination ticket ready to go making sure we track this for this week. 👍 from me I haven't been in loop aton on design for this change probaby have @malandis do a final pass also. Nothing standing out to me though in this PR. LGTM.

proto/cacheclient.proto
```diff
@@ -466,62 +466,82 @@ message _SortedSetPutRequest {
 
 message _SortedSetPutResponse {}
 
+message _Unbounded {}
+
```
wandaitzuchen, line 470: should use the `_Unbounded` message in the ListFetch fetch PR https://github.com/momentohq/client-protos/pull/126

malandis (approved):
Great!

PR #126, opened 2023-02-22
chore: Update ListFetch/ListRetain from count to end_index

ben-kugler:


proto/cacheclient.proto
```diff
@@ -402,16 +402,89 @@ message _ListRemoveRequest {
...
+
+  # start=None, end=None
+  list_[:] == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+
+  # start=5, end=None
+  list_[5:] == [5, 6, 7, 8, 9]
+
+  # start=5, end=0
+  list_[5:0] == [5, 6, 7, 8]
```
malandis, line 425: start=5, end=0, and forward iterating (step=1) means this evaluates to empty list, right? Ie this should be end=-1 to exclude the final list element

proto/cacheclient.proto
```diff
@@ -402,16 +402,89 @@ message _ListRemoveRequest {
...
+
+  # start=-5, end=None
+  list_[-5:] == [5, 6, 7, 8, 9]
+
+  # start=-5, end=0
+  list_[-5:0] == [5, 6, 7, 8]
```
malandis, line 431: Same thing here: start=-5, end=-1 (up to but not including the final list element)

malandis (commented):
Just a couple questions where end was zero

proto/cacheclient.proto
ben-kugler: Yep, you're totally right, corrected

proto/cacheclient.proto
```diff
@@ -402,16 +402,89 @@ message _ListRemoveRequest {
...
+// 2. `start_index` == 0, then we start at the front of list
+// 2. `end_index` == None, start at the end of the index, inclusive
+/*
+  list_ = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+  # Using the slice notation:
+  # start:stop(:step)?
+
+
+  # start=None, end=None
```
kvcache, line 417: nit: None == 0, I'd recommend just using `0` everywhere in here instead of also using the alias "none"

ben-kugler: Besides end_index, yeah? Start index can all just be 0, but for end index there's a difference between 0 and None, 0 being start of array and None being end of list.

for example, we have
```
  # start=None, end=None
  list_[:] == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```
and then 5:0
```
# start=5, end=0
  list_[5:0] == []
```

but what about, 0:0
```
# start=0, end=0
  list_[0:0] == []
```
or do we change end index being exclusive to inclusive for 0 values?
```
# start=0, end=0
  list_[0:0] == [0]
```

ben-kugler: Actually, this one case, 0,0 is a real problem. Since we treat the end index as exclusive, our logic says
```
# start=0, end=0
  list_[0:0] == []
```

but redis says
```
# start=0, end=0
  list_[0:0] == [0]
```

so how are we suppose to model this in ms2? Only options here is to,
1. In our docs we say that end_index is exclusive, unless 0 or omitted, if 0, then it includes the beginning of the list, if omitted, goes to end of list, this option actually makes sense and keeps it consistent with how we handle start/end list boundaries
2. in ms2, if we meet the case where start and end index is 0, increment start_index so that start is greater then end, causing redis to return a empty array [] 😩 👎 
```
# start=0, end=0
  list_[1:0] == []
```

ben-kugler: @malandis thoughts?

proto/cacheclient.proto
```diff
@@ -402,16 +402,93 @@ message _ListRemoveRequest {
...
   sint32 start_index = 2;
-  // If unset or 0, this effectively means list.length()
+  // Exclusive if set.
+  // If unset, inclusive, this effectively means list.length()
   // If count is > the number of elements to return, return as much as you can
-  uint32 count = 3;
+  oneof end_index {
+    Unbounded unbounded = 3;
+    sint32 inclusive = 4;
+  }
```
malandis, line 476: Looks good. We'll want the same pattern for `start_index`.

ben-kugler: Not necessary, the default value is 0 if `start_index` isn't included in the request, which is what we want. 

Start index can all just be 0, but for end index there's a difference between 0 and None, 0 being start of array and None being end of list.

malandis: For future compatibility, if we support reverse order, then an omitted `start_index` will mean "end of the list". The client library can set `start_index` to -1 in that case (end of list inclusive), but it's more logic to put in the clients.

malandis (approved):
Looks great!

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
+
+  # start=-5, end=0
+  list_[-5:0] == []
+
+
+  # start=0, end=5
+  list_[0:5] == [0, 1, 2, 3, 4]
+
+  # start=0, end=5
+  list_[0:5] == [0, 1, 2, 3, 4]
```
kvcache, line 438: these are copypasted, and if we're going to use inclusive numbers for end is this actually true?

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
+
+  # start=0, end=-5
+  list_[0:-5] == [0, 1, 2, 3, 4]
+
+  # start=0, end=-5
+  list_[0:-5] == [0, 1, 2, 3, 4]
```
kvcache, line 444: actually looks like the same 2 issues as above here

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
+  list_[0:] == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+
+  # start=5, end=None
+  list_[5:] == [5, 6, 7, 8, 9]
+
+  # start=5, end=0
+  list_[5:0] == []
+
+  # start=-5, end=None
```
kvcache, line 427: I think `None` might be better documented as `Unbounded` now that you've got a first-class definition for that notion?

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
+  list_[-7:-2] == [3, 4, 5, 6, 7]
+
+  // if using ltrim, this will remove all elements in the list
+  # start=-2, end=-6
+  list_[2:-6] == []
+
+  # empty list
+  list_[-1:1] == []
+ */
+message Unbounded { }
```
wandaitzuchen, line 463: `_Unbounded`?

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
+
 message _ListFetchRequest {
   bytes list_name = 1;
   // Inclusive. 0 (start of list) by default.
```
wandaitzuchen, line 467: did you mean we use 0 if unbounded?

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
   // A negative index counts from the end of the list
-  sint32 start_index = 2;
-  // If unset or 0, this effectively means list.length()
+  oneof start_index {
+    Unbounded unbounded_start = 2;
+    sint32 inclusive_start = 3;
+  }
+  // Exclusive if set.
+  // If unset, inclusive, this effectively means list.length()
   // If count is > the number of elements to return, return as much as you can
```
wandaitzuchen, line 475: are these comments outdated?

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
 message _ListFetchRequest {
   bytes list_name = 1;
   // Inclusive. 0 (start of list) by default.
   // A negative index counts from the end of the list
-  sint32 start_index = 2;
-  // If unset or 0, this effectively means list.length()
+  oneof start_index {
+    Unbounded unbounded_start = 2;
+    sint32 inclusive_start = 3;
+  }
```
kvcache, line 472: does `0` mean the same thing as `Unbounded`? Should we consider not using Unbounded for start indices?

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
+  // Exclusive if set.
+  // If unset, inclusive, this effectively means list.length()
```
kvcache, line 474: How can someone set this oneof to something exclusive? Both options are inclusive. Should this comment exist?

kvcache (commented):
Okay, here's a little more detail about the things I was trying to allude to 😄

proto/cacheclient.proto
kvcache: Oh yeah, nice catch Wanda

proto/cacheclient.proto
```diff
@@ -402,16 +402,99 @@ message _ListRemoveRequest {
...
+    sint32 inclusive_end = 5;
+  }
+}
+
+message _ListRetainRequest {
+  bytes list_name = 1;
+  oneof start_index {
+    Unbounded unbounded_start = 2;
+    sint32 inclusive_start = 3;
+  }
```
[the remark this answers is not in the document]
ben-kugler: Yep, unfortunately, the thread was folded, but `start_index` will have a explicit presence so that we can support reverse order in the future

proto/cacheclient.proto
ben-kugler: Most of this is out of date, rewriting

proto/cacheclient.proto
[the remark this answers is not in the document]
ben-kugler: So, we would be adding in the future, if asked, a numerical variable would be added, lets use `step` for now, this would control our traversal direction as well as controlling our stride, so not just reversing the list.

This would be a optional param and would not change any of the default behavior for existing workflows. But if `step` is included in the request and is negative and `start` is Unbounded, then `starts` becomes (list.len - 1)

But overall, I don't think making `start`  have a explicit presence hurt us, it might add to confusion as to why, but documentation here can cover that gap.

@malandis, anything to add? Also to double check what I got here is right 😄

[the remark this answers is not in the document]
kvcache: ...i see.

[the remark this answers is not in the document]
malandis: Yes that's right. In our zoom conversation we talked about the same thing but with start, count, and "direction of travel" (if I remember that right?). The `start_index` behavior we're talking about here will support that should we 
 implement it

proto/cacheclient.proto
```diff
@@ -402,16 +402,110 @@ message _ListRemoveRequest {
...
+    _Unbounded unbounded_start = 2;
+    sint32 inclusive_start = 3;
+  }
+  // Exclusive.
+  // If unbounded, this effectively means list.length()
+  // If end_index is > the number of elements to return, return as much as you can
+  // A negative index counts from the end of the list
+  oneof end_index {
+    _Unbounded unbounded_end = 4;
+    sint32 inclusive_end = 5;
```
wandaitzuchen, line 489: should call this exclusive_end then?

ben-kugler: Oh, yes, thank you

proto/cacheclient.proto
```diff
@@ -402,16 +402,110 @@ message _ListRemoveRequest {
...
+
+  # start and end are zero-based indexes, both are optional
+
+  # start is inclusive
+  # If start is not set, then the default is Unbounded, which is treated as 0
+  # Unbounded == 0,
+  # this field has a explicit presence for future compatibility to support reverse order and controlling our stride
+
+  # end is exclusive
+  # If end is not set, then the default is Unbounded which becomes (list.len)
```
kvcache, line 416: great, this is clear and I'm glad you're using half-open now!

proto/cacheclient.proto
```diff
@@ -402,16 +402,110 @@ message _ListRemoveRequest {
...
+  list_[5:] == [5, 6, 7, 8, 9]
+
+  # start=5, end=0
+  list_[5:0] == []
+
+  # start=-5, end=Unbounded
+  list_[-5:] == [5, 6, 7, 8, 9]
+
+  # start=-5, end=0
+  list_[-5:0] == []
```
malandis, line 442: The example directly above is a duplicate of 432-433
```suggestion
```

proto/cacheclient.proto
```diff
@@ -402,16 +402,110 @@ message _ListRemoveRequest {
...
+  # Unbounded == 0,
+  # this field has a explicit presence for future compatibility to support reverse order and controlling our stride
+
+  # end is exclusive
+  # If end is not set, then the default is Unbounded which becomes (list.len)
+
+  # start and end can be negative numbers, negative indicates offsets starting at the end of the list.
+  # For example, -1 is the last element of the list, -2 the penultimate, etc
+  #               [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+  # +start/+end    0, 1, 2, 3, 4, 5, 6, 7, 8, 9
```
kvcache, line 421: What does "+start/+end" mean? Does it mean `0/9`? That would output 0,1...,7,8 though based on the preceding rule.

Oh.. do you mean
```suggestion
  # For example, -1 is the last element of the list, -2 the penultimate, etc
  # list:                     [  a,  b,  c,  d,  e,  f,  g,  h,  i,  j]
  # positive index positions:    0,  1,  2,  3,  4,  5,  6,  7,  8,  9
  # negative index positions:  -10, -9, -8, -7, -6, -5, -4, -3, -2, -1
```

proto/cacheclient.proto
```diff
@@ -402,16 +402,110 @@ message _ListRemoveRequest {
...
+
+  # start=2, end=-6
+  list_[2:-6] == [2, 3]
+
+  # start=-7, end=-2
+  list_[-7:-2] == [3, 4, 5, 6, 7]
+
+  // if using ltrim, this will remove all elements in the list
+  # start=-2, end=-6
+  list_[2:-6] == []
```
kvcache, line 467: Why should we interpret an empty range that _looks_ like a partial deletion was _intended_ in this 1 special case as "delete your entire list" instead of the more conservative "do nothing?"
Isn't this inconsistent with the fetch behavior, where that same partial range intent doesn't end up mean "return me the whole list?

This is a footgun, we should talk about it before moving ahead.

proto/cacheclient.proto
```diff
@@ -402,16 +402,110 @@ message _ListRemoveRequest {
...
+  # start=2, end=6
+  list_[2:6] == [2, 3, 4, 5]
+
+  # start=2, end=-6
+  list_[2:-6] == [2, 3]
+
+  # start=-7, end=-2
+  list_[-7:-2] == [3, 4, 5, 6, 7]
+
+  // if using ltrim, this will remove all elements in the list
```
kvcache, line 465: We don't have ltrim, we have ListRetain right?

wandaitzuchen:
LGTM! Will wait for ✅ from @malandis.

proto/cacheclient.proto
```diff
@@ -402,16 +402,110 @@ message _ListRemoveRequest {
...
+  # start=-2, end=-6
+  list_[2:-6] == []
+
+  # empty list
+  list_[-1:1] == []
+ */
```
kvcache, line 471: I think this is wrong / only true on empty lists and lists greater than length 1. Consider a list containing `[a]`. What is `[-1:1]`? My brain says it would be `[a]` and Python agrees it's `['a']`.
```
$ python3 -c 'print(["a"][-1:1])'
['a']
```

I assert the correct empty range definition is actually `[0:0]`. That directly implies that the exclusivity at the end of the range needs to take precedence over the inclusivity at the start of the range.

Can you provide a set of expectations for lists of length 1 with respect to what special range cases mean? I'm not sure if lists of length 2 will also have special considerations.

kvcache (commented):
Okay, the intent is much clearer - I spy a couple gaps but we're getting there!

proto/cacheclient.proto
ben-kugler: I agree, I think if you bounds are invalid, then we shouldn't delete the entire list. (also, the comment says start=-1, end=-6, but notation says [2:-6], notation here is wrong.) 

Previously, we styling after redis implementation of of ltrim, invalid bounds would cause the list to be removed/emptied, I would lean more towards returning a invalid bounds response, how do we feel about

```
message _ListRetainResponse {
  message _Success {}

  message _InvalidBounds {}

  oneof list {
    _Success success = 1;
    _InvalidBounds invalid = 2;
  }
}
```?

File: proto/cacheclient.proto (hunk 20 above)
Comment (malandis) [reply proto/cacheclient.proto:474]: Agreed. The other convention to document is that we implicitly clip out of range end indexes to `list.len`. Ie in this case there is no list element with index `1`, so end=1 implies `end=Unbounded`

Comment (ben-kugler) [reply proto/cacheclient.proto:474]: Oh good catch.

Looks like list of length 2 aren't affected by the same special cases

File: proto/cacheclient.proto (hunk 18 above)
Comment (kvcache) [reply proto/cacheclient.proto:470]: Oh interesting, I don't have a strong opinion about whether to prefer [] or _InvalidBounds. It would probably be nicer for ListRetain to let you know you tried to do something that can't ever do anything. But then I wonder if it should be done for fetch as well?

File: proto/cacheclient.proto [hunk 21]
```diff
@@ -402,16 +402,113 @@ message _ListRemoveRequest {
...
+  # If end is not set, then the default is Unbounded which becomes (list.len)
+
+  # start and end can be negative numbers, negative indicates offsets starting at the end of the list.
+  # For example, -1 is the last element of the list, -2 the penultimate, etc
+  #               [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+  # For example, -1 is the last element of the list, -2 the penultimate, etc
+  # list:                     [  a,  b,  c,  d,  e,  f,  g,  h,  i,  j]
+  # positive index positions:    0,  1,  2,  3,  4,  5,  6,  7,  8,  9
+  # negative index positions:  -10, -9, -8, -7, -6, -5, -4, -3, -2, -1
+  # -start/-end  -10,-9,-8,-7,-6,-5,-4,-3,-2,-1
```
kvcache, line 425: formatting got pasted wonky

proto/cacheclient.proto
ben-kugler: But, to clarify, this would only be for invalid bounds, wouldn't work for bounds that are valid, but out of range of your list. You can try and retain out of bounds, but that would mean trimming the entire list.           
```
list - [0,1,2,3,4,5,6,7,8,9]
ListRetain(start_index: 100, end_index: 102)
list - []
```

Yeah, not opposed to adding to ListFetch

malandis: I can see a case for an error for `ListRetain`, ie if they want to delete the whole thing they should do that. But for `ListFetch` I prefer returning an empty slice as that is a convention familiar in various languages.

Should we go with a gRPC error, I prefer `OUT_OF_RANGE` (see [here](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)) as that maps cleanly to our use case and is otherwise unambiguous. For other list indexing operations (eg "get the value at position 10") this is appropriate as well.

malandis (dismissed):
LGTM!

proto/cacheclient.proto
```diff
@@ -402,16 +402,125 @@ message _ListRemoveRequest {
...
+
+  # start=2, end=-6
+  list_[2:-6] == [2, 3]
+
+  # start=-7, end=-2
+  list_[-7:-2] == [3, 4, 5, 6, 7]
+
+  // if using ListRetain, this will remove all elements in the list
+  # start=-2, end=-6
+  list_[2:-6] == []
```
wandaitzuchen, line 467: `list_[-2:-6]`?

proto/cacheclient.proto
```diff
@@ -402,16 +402,125 @@ message _ListRemoveRequest {
...
+  list_size_one_[-1:1] == ['a']
+
+  # start=1, end=-1
+  list_size_one_[1:-1] == []
+
+  # start=0, end=0
+  list_size_one_[0:0] == []
+
+  # start=0, end=-1
+  list_size_one_[-1:0] == []
```
wandaitzuchen, line 485: `list_size_one_[0:-1] == []`?

proto/cacheclient.proto
```diff
@@ -402,16 +402,125 @@ message _ListRemoveRequest {
...
+ */
+message _Unbounded { }
+
```
wandaitzuchen, line 488: Could you merge from main and clean up https://github.com/momentohq/client-protos/blob/main/proto/cacheclient.proto#L469 ? Also you'll have to build again with what's in main to get `client_protos.dsc` correctly updated.

wandaitzuchen (approved):
🚢

malandis (approved):
💥

PR #127, opened 2023-02-27
feat: use oneof to determine if token has expiry

bruuuuuuuce:
After a back and forth, we have decided to use oneof when generating api tokens, to determine if a token expires or not. Previously, we were going to have `valid_for_seconds = 0` to mean that the token would never expire. This change just makes it much more explicit

### DIFF
--- a/proto/controlclient.proto
+++ b/proto/controlclient.proto
@@ -104,11 +104,13 @@ message _GenerateTokenRequest {
 }
 
 message _GenerateTokenResponse {
-  // the base64 encoded (api_key, endpoint)
+  // the token that will be used to make requests against Momento backend
   string api_token = 1;
   // the token that will allow the api token to be refreshed, which will
   // give you back a new refresh and api token
   string refresh_token = 2;
+  // the Momento endpoint that this token is allowed to make requests against
+  string endpoint = 3;
   // epoch seconds when the api token expires
-  uint64 valid_until = 3;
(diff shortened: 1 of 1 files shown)


rlinehan (dismissed):
Looks good to me! I think getting Chris' or someone else on dev-eco's approval would be good though.

kvcache (approved):
Looks fine, and explicit is better than implicit.

proto/controlclient.proto
```diff
@@ -86,14 +86,26 @@ message _FlushCacheResponse {
...
+    Never never = 2;
+    Expires expires = 3;
+  }
+
+  reserved 1;
 }
 
-message _GenerateTokenResponse {
+message _GenerateApiTokenResponse {
   // the token that will be used to make requests against Momento backend
```
rlinehan, line 107: What do you think about updating this slightly to say like "the api key used for authentication against Momento backend" or something, to try to call out that it's just the bit used for authentication?

rlinehan (approved):
Looks good! Left one comment about a comment (😅) but not blocking.

proto/controlclient.proto
bruuuuuuuce: fixed in https://github.com/momentohq/client-protos/pull/127/commits/56d96792c8d8a421a528caf9168c801351a60e7f

rlinehan (approved):
:shipit: 🚀

PR #131, opened 2023-03-01
fix: Add missing go_package options.

schwern:
A couple proto files were mising the `go_package` option. We keep having to put it back when copying the protos to Go.

### DIFF
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -1,5 +1,6 @@
 syntax = "proto3";
 
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
 option java_multiple_files = true;
 option java_package = "momento.auth";
 
--- a/proto/cacheping.proto
+++ b/proto/cacheping.proto
@@ -1,5 +1,6 @@
 syntax = "proto3";
 
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
 option java_multiple_files = true;
 option java_package = "grpc.cache_client";
 option csharp_namespace = "Momento.Protos.CachePing";


schwern:
Without those lines we get...

```
protoc -I=internal --go_out=internal --go_opt=paths=source_relative --go-grpc_out=internal --go-grpc_opt=paths=source_relative internal/protos/*.proto
protoc-gen-go: unable to determine Go import path for "protos/auth.proto"

Please specify either:
	• a "go_package" option in the .proto source file, or
	• a "M" argument on the command line.

See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.

--go_out: protoc-gen-go: Plugin failed with status code 1.
make: *** [protos] Error 1
```

rlinehan:
> The Go import path may be specified on the command line when invoking the compiler, by passing one or more M${PROTO_FILE}=${GO_IMPORT_PATH} flags. Example usage:
> 
> protoc --proto_path=src \
>   --go_opt=Mprotos/buzz.proto=example.com/project/protos/fizz \
>   --go_opt=Mprotos/bar.proto=example.com/project/protos/foo \
>   protos/buzz.proto protos/bar.proto
> 
> Since the mapping of all .proto files to their Go import paths can be quite large, this mode of specifying the Go import paths is generally performed by some build tool (e.g., [Bazel](https://bazel.build/)) that has control over the entire dependency tree. If there are duplicate entries for a given .proto file, then the last one specified takes precedence.
> 
> For both the go_package option and the M flag, the value may include an explicit package name separated from the import path by a semicolon. For example: "example.com/protos/foo;package_name". This usage is discouraged since the package name will be derived by default from the import path in a reasonable manner.

That's what the doc linked (https://developers.google.com/protocol-buffers/docs/reference/go-generated#package) says. I wonder if we should be doing this in whatever we have that generates our go stuff instead?

schwern:
@rlinehan That would complicate generating the Go protos, which we [currently do manually in a Makefile](https://github.com/momentohq/client-sdk-go/blob/accb869b6e7bdcbc585dcd86ec03aa1e2790feb3/Makefile#L36). We wouldn't be able to just do `*.proto`, we'd need to add a -M for every file.

The other proto files already have the go_package option, so this is just making them all behave the same.

Eventually we'll properly package them.

kvcache (approved):

PR #132, opened 2023-03-01
chore: Adding protos for KeyExists

mjindalmotiv8:
Adds protocols for a KeyExists API, which can be used to query whether keys exist or not

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -21,6 +21,7 @@ service Scs {
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {}
   rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
+  rpc KeysExists (_KeysExistsRequest) returns (_KeysExistsResponse) {}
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -21,6 +21,7 @@ service Scs {
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {}
   rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
+  rpc KeysExists (_KeysExistsRequest) returns (_KeysExistsResponse) {}
```
kvcache, line 24:
```suggestion
  rpc KeysExist (_KeysExistRequest) returns (_KeysExistResponse) {}
```
for it to read a little more naturally (especially for the poor enunciators among us)?

kvcache (commented):
looks good, 1 very small question

proto/cacheclient.proto
mjindalmotiv8: I actually had made this change and updated the doc - did not push the commit yet. Updated just now

kvcache (approved):
l's gtm

PR #137, opened 2023-03-16
feat: add api for refreshing api tokens

bruuuuuuuce:
design doc https://www.notion.so/momentohq/New-API-Refresh-Api-Token-cfe33d84e2e24d409a6fdd05ea522ac9

### DIFF
--- a/proto/controlclient.proto
+++ b/proto/controlclient.proto
@@ -17,6 +17,7 @@ service ScsControl {
   rpc ListSigningKeys (_ListSigningKeysRequest) returns (_ListSigningKeysResponse) {}
   // api for programatically generating api and refresh tokens
   rpc GenerateApiToken(_GenerateApiTokenRequest) returns (_GenerateApiTokenResponse) {}
+  rpc RefreshApiToken(_RefreshApiTokenRequest) returns (_RefreshApiTokenResponse) {}
 }
 
 message _DeleteCacheRequest {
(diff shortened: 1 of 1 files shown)


proto/controlclient.proto
```diff
@@ -114,3 +115,20 @@ message _GenerateApiTokenResponse {
   // epoch seconds when the api token expires
   uint64 valid_until = 4;
 }
+
+message _RefreshApiTokenRequest {
+  string api_token = 1;
+	string refresh_token = 2;
```
bruuuuuuuce, line 121:
```suggestion
string refresh_token = 2;
```

proto/controlclient.proto
```diff
@@ -114,3 +115,20 @@ message _GenerateApiTokenResponse {
   // epoch seconds when the api token expires
   uint64 valid_until = 4;
 }
+
+message _RefreshApiTokenRequest {
+  string api_token = 1;
+  string refresh_token = 2;
+}
```
rlinehan, line 122: Ergh, I should have thought of this in the design meeting (although this is less a UX design thing and more an implementation detail)... it might be worth doing some testing of how we want the control plane code to look before merging this. I think there's an open question of should the api token come through in a header, like it normally does, and then we do our normal validation of it except it's ok if it's expired, or do we want to have the endpoint be completely unauthenticated and check this as a parameter?

I'm leaning a bit towards it comes through in the header, because that seems easier for the SDK to do (i.e. that's already part of the client configuration)? If it does come through in the header then I don't think we need it here.

bruuuuuuuce: ah yea thats a good call. Let me play around with it a bit and get back to you here

PR #138, opened 2023-03-16
feat: Providing list length info for PopFront, PopBack, Remove, Retain, and Erase Responses.

honeyAcorn:
[Problem]
Clients may get list length after list modification command; resulting in race-condition. 
For more information, please refer to:
https://github.com/momentohq/cacheservice/issues/1197

[Solution]
Including list length in PopFront, PopBack, Remove, Retain, Erase responses.


wandaitzuchen (approved):
LGTM! I'll wait for @cprice404 to take a look before merging.

proto/cacheclient.proto
```diff
@@ -402,6 +402,8 @@ message _ListPopFrontRequest {
 message _ListPopFrontResponse {
   message _Found {
     bytes front = 1;
+    // length of the list after the pop
```
mjindalmotiv8, line 405: @cprice404 - can you confirm that these responses will not break SDK's?

cprice404: I'm 90% sure that this is fine.  They will just not be deserialized by the SDKs until we update the SDK protos.

But, to be safe, we should just test by running integration tests in a dev cell and/or alpha when we have a working version of the corresponding ms2/mr2 changes.

cprice404 (approved):
These changes look fine to me but y'all should figure out an order of operations for how you want to test them with ms2+mr2+existing SDK integration tests and decide when to merge based on that.

wandaitzuchen:
@HoneyAcron please hold off merging this until later. We should test SDKs out when you have the backend implementations ready in your dev cell.

HoneyAcron:
>

HoneyAcron:
I will leave this open/hold for now until we have fully tested it with the backend implementation. I accidentally close this PR so now I'm reopening it.

mjindalmotiv8:
In accordance with Chris's comments - here is my feedback on how to go about this - leave this proto open

1. Work on MS2 and merge the MS2 change
2. Work on MR2 with local version of client proto
3. Test that SDK is not broken in dev cell
4. Commit this proto change if SDK works
5. Commit the MR2 change
6. Work on the SDK changes

HoneyAcron:
Added Kenny's feedback. For responses in Erase, Remove, Retain, we need to return Found or Missing and include the length in Found. This is because previously user don't need to know if the operation is successful. If successful, it will just delete/modify. However, if the key/value doesn't exist, it simply won't do anything, so no need to provide a response.

Now that we are responding with list_length, if user get a 0 for list length, they can't really tell if it's the real length or the modification failed. Thus adding new fields.

proto/cacheclient.proto
```diff
@@ -448,7 +452,18 @@ message _ListEraseRequest {
   }
 }
 
-message _ListEraseResponse {}
+message _ListEraseResponse {
+  message _Found {
```
cprice404, line 456: @kvcache same story here w/rt backward compat, right?  Existing SDKs will just ignore this data and continue to operate as they did before, yes?

proto/cacheclient.proto
```diff
@@ -458,7 +473,18 @@ message _ListRemoveRequest {
   }
 }
 
-message _ListRemoveResponse {}
+message _ListRemoveResponse {
+  message _Found {
+    uint32 list_length = 1;
+  }
+
+  message _Missing {}
```
cprice404, line 481: this stuff is going to be a little more interesting w/rt the SDKs.  Today, we have several 1.0 SDKs that include a ListRemoveValue function, and they currently have only `Success` and `Failure` response types.

It's easy enough to add the new list length value to the `Success` type, but changing the types be a breaking change.  We could theoretically do something like add a `Miss` response type and leave `Success` in place, mapped to `_Found`, but that would be a little inconsistent with our existing APIs.

@malandis @kvcache I'm interested in thoughts on this.

kvcache: I think it's not a huge issue to expose found/missing through success to the user. In fact, I recommend mapping it to `Success { list_length: 0 }` on the SDK Success object to start out with, and when requested you can add a method `fn was_found() -> bool` (or whatever name you like) to Success without breaking anyone.

cprice404: or I guess `listLength` could be of type `int?` ?

cprice404:
If @malandis is on board with condensing Found/Missing => Success in the sdks then I'm +1 on this

proto/cacheclient.proto
kvcache: You're talking about `momento_sdk::Response::Success::listLength`? You could do that too. Imo it's more annoying than letting it be `int` and having to ask for "did it do the delete?" separately. I'd hazard that the 95% use case is only concerned with `int`. As far as the protos go, I'm just trying to help keep a non-fire-drill avenue open for helping the sdk's serve the remaining vocal 5%, should they speak up.

cprice404: yeah you're right re: 90% case.

honeyAcorn:
@cprice404 I've tested my [MR2](https://github.com/momentohq/cacheservice/pull/1318) and [MS2](https://github.com/momentohq/storage-store/pull/176) code changes with the client-sdk-dotnet integration test and it's backward compatible. Is there any other sdk that you think we can test it with before we merge this PR?

cprice404 (approved):

PR #139, opened 2023-03-20
feat: new service for protos with unique auth methods

bruuuuuuuce:
We have found that apis that require non-standard auth methods don't fit well into our current services. This pr is an attempt to make a new service which contains these apis. For now, the only 2 methods that fit this use case are `GenerateApiToken` and `RefreshApiToken`. We could also loop these into the `Auth` service proto, instead of creating a new `Token` service

https://github.com/momentohq/client-protos/blob/main/proto/auth.proto


rlinehan:
We're going to do #140 .

PR #141, opened 2023-03-20
feat: make java protos publishable to maven central

nand4011:
Add the maven publish plugin and nexus publish plugin to make the java protos publishable to maven central.

Generate javadoc and sources jars based on the generated code.

Add the dependencies plugin to force non-transitive dependencies.


java/build.gradle.kts
```diff
@@ -49,3 +77,49 @@ protobuf {
...
+                licenses {
+                    license {
+                        name.set("The Apache License, Version 2.0")
+                        url.set("https://www.apache.org/licenses/LICENSE-2.0.txt")
+                    }
+                }
+                developers {
+                    developer {
+                        id.set("nand4011")
+                        name.set("Nate Anderson")
```
cprice404, line 100: i'm okay with this for now if you are!

java/build.gradle.kts
```diff
@@ -7,22 +7,50 @@ import com.google.protobuf.gradle.protoc
...
 dependencies {
     implementation("io.grpc:grpc-protobuf:$grpcVersion")
+    implementation("com.google.api.grpc:proto-google-common-protos:2.0.1")
+
+    implementation("com.google.protobuf:protobuf-java:${grpcProtobufVersion}")
+    implementation("com.google.guava:guava:${guavaVersion}")
+
     implementation("io.grpc:grpc-stub:${grpcVersion}")
-    implementation("javax.annotation:javax.annotation-api:1.3.2")
+    implementation("io.grpc:grpc-api:${grpcVersion}")
```
cprice404, line 30: just curious why it looks like there are a bunch of new grpc deps coming in.  Are these just transitive deps from grpc-stub?

nand4011: Yeah. Those are all the formerly transitive dependencies being used.

.github/workflows/push.yml
```diff
@@ -116,6 +116,26 @@ jobs:
...
+      - name: publish to sonatype
+        env:
+          SONATYPE_SIGNING_KEY: ${{ secrets.SONATYPE_SIGNING_KEY }}
+          SONATYPE_SIGNING_KEY_PASSWORD: ${{ secrets.SONATYPE_SIGNING_KEY_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          VERSION: ${{ needs.release.outputs.version }}
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: publishToSonatype closeSonatypeStagingRepository  #TODO: closeAndReleaseStagingRepository
```
nand4011, line 137: This won't release the artifact automatically. We can start doing that once we manually release the first working version.

java/build.gradle.kts
```diff
@@ -7,22 +7,50 @@ import com.google.protobuf.gradle.protoc
...
+            "build/generated/source/proto/main/java",
+    )
+}
+
+java {
+    withJavadocJar()
+    withSourcesJar()
+    // Produce an artifact build for Java 8+
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
```
nand4011, line 50: I was pretty conservative with the language version here. Should we have a higher minimum?

cprice404: i think this is right for client-protos.  we might have a discussion about the SDK itself but for client-protos let's stick with 8.

cprice404 (commented):
just one q about the env var name

java/build.gradle.kts
```diff
@@ -49,3 +77,61 @@ protobuf {
         }
     }
 }
+
+publishing {
+    publications {
+        create<MavenPublication>("mavenJava") {
+            from(components["java"])
+            version = System.getenv("JAVA_PROTOS_VERSION")
```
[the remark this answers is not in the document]
nand4011: Good catch

cprice404 (approved):

PR #146, opened 2023-03-22
chore: auto-release the java artifact and update dependencies

nand4011:
Automatically release the artifact when the publish_java workflow runs.

Switch to gRPC and Protobuf BOMs for dependency versioning and update to the latest versions.

Note which dependency versions are pulled from other dependencies for future updates.

When I get the kotlin protos publishing to maven central I'll factor the gRPC/Protobuf dependencies to a common catalog.


java/generated-sources/descriptors/client_protos.dsc
nand4011, line 1: I don't know what this file does. I had no issues using the new dependency versions in the java sdk.

cprice404: @kvcache do you know what this file is?  should it be in VCS?

java/build.gradle.kts
```diff
@@ -19,19 +19,21 @@ plugins {
...
-val grpcProtobufVersion = "3.21.2"
-val grpcVersion = "1.47.0"
-val guavaVersion = "31.1-android"
+val grpcProtobufVersion = "3.22.2"
+val grpcVersion = "1.53.0"
 
 dependencies {
-    implementation("io.grpc:grpc-protobuf:$grpcVersion")
-    implementation("com.google.api.grpc:proto-google-common-protos:2.0.1")
+    api(platform("io.grpc:grpc-bom:$grpcVersion"))
```
cprice404, line 26: nice, thanks for finding the bom.

does this need to be an `api` dep though?

java/build.gradle.kts
```diff
@@ -19,19 +19,21 @@ plugins {
...
+    api(platform("com.google.protobuf:protobuf-bom:$grpcProtobufVersion"))
 
-    implementation("com.google.protobuf:protobuf-java:${grpcProtobufVersion}")
-    implementation("com.google.guava:guava:${guavaVersion}")
+    api("io.grpc:grpc-stub")
+    api("io.grpc:grpc-api")
+    api("com.google.protobuf:protobuf-java")
+    api("com.google.guava:guava:31.1-android") // version pulled from protobuf-java
```
cprice404, line 32: So are you switching these from impl to `api` because there are types from them that appear in the APIs of the generated code?  And thus one can't consume the generated code without these deps?

Mostly asking because of concern about possibly conflicting versions of these between us and our users.

java/build.gradle.kts
nand4011: Gradle uses both api and implementation in their documentation. I used api because some of the dependencies are api dependencies. I'll try again to find out what api vs implementation means for a bom.

java/build.gradle.kts
nand4011: Yep. grpc-stub, protobuf-java, and guava are required to use the apis. I don't think grpc-api is, so I'll change it. I really want a tool that lets me see every place a dependency is used in a project.

nand4011: The presence of guava does give me a bit of pause. I want to investigate shading this and many of the other dependencies in the java sdk.

java/build.gradle.kts
```diff
@@ -19,19 +19,21 @@ plugins {
...
+
+    implementation("io.grpc:grpc-protobuf")
+    implementation("com.google.api.grpc:proto-google-common-protos:2.9.0") // version pulled from grpc-protobuf
```
kvcache, line 35: this is unnecessary, is it not?

java/generated-sources/descriptors/client_protos.dsc
kvcache: I do know what it is and I'd recommend against including it in VCS for a public client artifact. It should be useless to java clients.

java/build.gradle.kts
nand4011: The transitive dependency analysis plugin flags it if it isn't specified. Is this only used during the generation process? It could just be confusing the plugin.

java/build.gradle.kts
nand4011: Do we want to leave them as implementation to force users to explicitly depend on them? The former submodule in the java sdk used api dependencies transitively before I made them explicit.

cprice404 (approved):

PR #150, opened 2023-03-31
feat: Move GenerateApiToken to auth proto

rlinehan:
Previously the GenerateApiToken API was in `controlclient.proto`. However, this API is going to require some special auth considerations. On the server side, we wrap each grpc service in a `UserHeaderInterceptor` - each API goes through this first, and there isn't a good way to determine which API is being called within it, so doing special case auth is really challenging.

Instead, this moves the API into `auth.proto`. Currently this service just hosts the Login action, which is unauthenticated. Not only does `auth` seem a logical place to put this semantically, but implementation-wise each API here can handle it's own authentication because we just won't have a broad header interceptor across all of them.

This is a breaking change. However, we haven't fully "released" anything using this API yet (there's a `preview` CLI command we'll need to update, but that hasn't been usable externally anyway).


ben-kugler (approved):

bruuuuuuuce (approved):
:shipit:

PR #151, opened 2023-04-03
feat: Specify whether cache methods are idempotent

nand4011:
Use the idempotency_level option to specify whether the cache methods are safe, idempotent, or unknown. Safe methods have no side effects and are implicitly idempotent. Safe and idempotent methods can be retried.

We currently specify idempotency per-sdk using a manually updated list:
[.NET](https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs#L40)
[node.js](https://github.com/momentohq/client-sdk-nodejs/blob/main/src/config/retry/default-eligibility-strategy.ts#L31)

For grpc implementations that support it, we can check whether a method is idempotent in the interceptor. If a grpc implementation doesn't, this will still function as the source of truth for whether a method is idempotent.

I tested that this works using the Java integration tests and by checking idempotency in an existing interceptor.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -17,36 +17,90 @@ enum ECacheResult {
 }
 
 service Scs {
-  rpc Get (_GetRequest) returns (_GetResponse) {}
-  rpc Set (_SetRequest) returns (_SetResponse) {}
-  rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {}
-  rpc Delete (_DeleteRequest) returns (_DeleteResponse) {}
-  rpc KeysExist (_KeysExistRequest) returns (_KeysExistResponse) {}
-  rpc Increment (_IncrementRequest) returns (_IncrementResponse) {}
-  rpc UpdateTtl (_UpdateTtlRequest) returns (_UpdateTtlResponse) {}
+  rpc Get (_GetRequest) returns (_GetResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  rpc Set (_SetRequest) returns (_SetResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
+  rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
+  rpc Delete (_DeleteRequest) returns (_DeleteResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
+  rpc KeysExist (_KeysExistRequest) returns (_KeysExistResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  rpc Increment (_IncrementRequest) returns (_IncrementResponse) {
+    option idempotency_level = IDEMPOTENCY_UNKNOWN;
+  }
+  rpc UpdateTtl (_UpdateTtlRequest) returns (_UpdateTtlResponse) {
+    option idempotency_level = IDEMPOTENCY_UNKNOWN;
+  }
 
-  rpc DictionaryGet (_DictionaryGetRequest) returns (_DictionaryGetResponse) {}
-  rpc DictionaryFetch (_DictionaryFetchRequest) returns (_DictionaryFetchResponse) {}
-  rpc DictionarySet (_DictionarySetRequest) returns (_DictionarySetResponse) {}
-  rpc DictionaryIncrement (_DictionaryIncrementRequest) returns (_DictionaryIncrementResponse) {}
-  rpc DictionaryDelete (_DictionaryDeleteRequest) returns (_DictionaryDeleteResponse) {}
+  rpc DictionaryGet (_DictionaryGetRequest) returns (_DictionaryGetResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  rpc DictionaryFetch (_DictionaryFetchRequest) returns (_DictionaryFetchResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
nand4011, line 1: I chose how to mark the methods based on the list in the .NET sdk and educated guesses. We need someone with more familiarity than me to go over these before this can go in.

proto/cacheclient.proto
```diff
@@ -17,36 +17,90 @@ enum ECacheResult {
...
+  }
+  rpc KeysExist (_KeysExistRequest) returns (_KeysExistResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  rpc Increment (_IncrementRequest) returns (_IncrementResponse) {
+    option idempotency_level = IDEMPOTENCY_UNKNOWN;
+  }
+  rpc UpdateTtl (_UpdateTtlRequest) returns (_UpdateTtlResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
```
kvcache, line 40: this is not idempotent. Look at the update ttl request message

kvcache (commented):
Cool, I only see 1 miss

proto/cacheclient.proto
nand4011: I was wondering about this one. I assumed idempotent because I couldn't think of a case where a first try would not update the ttl but an additional try would. Setting it to the same ttl multiple times wouldn't be a problem.

kvcache: > Setting it to the same ttl multiple times wouldn't be a problem.

Can you explain the result of each of these if re-executed? https://github.com/momentohq/client-protos/blob/main/proto/cacheclient.proto#L146-L153 

fyi "wouldn't be a problem" is not a sufficient condition for an action to be correctly described as "idempotent." A quick google should help clear the meaning up!

nand4011: An idempotent function is one that has no further effect after the first call, or rather can be called any number of times with the same effect as being called one time. I would appreciate if you told me the specific problem you have instead of asking me to google word definitions.

If an object has a ttl of 1000 and I call this function to set it to 2000:
increase_to_milliseconds: The first call sets it to 2000 and subsequent calls leave it at 2000.
decrease_to_milliseconds: The first call leaves it at 1000 as do subsequent calls
overwrite_to_milliseconds: The first call sets it to 2000 and subsequent calls leave it at 2000.

If an object has a ttl of 3000 and I set it to 2000:
increase_to_milliseconds: The first call leaves the ttl at 3000 as do subsequent calls.
decrease_to_milliseconds: The first call sets it to 2000 and subsequent calls leave it at 2000.
overwrite_to_milliseconds: The first call sets it to 2000 and subsequent calls leave it at 2000.

I certainly could be missing a case or not considering side effects like the ttl resetting, but I don't immediately see it. What is your specific objection? How is setting a ttl to a specific number 5 times any different from setting it once?

kvcache: > An idempotent function is one that has no further effect after the first call, or rather can be called any number of times with the same effect as being called one time. I would appreciate if you told me the specific problem you have instead of asking me to google word definitions.

Cool, sounds like you understand the word now. Thanks. It's important to understand what you are PRing. In response to your normative request, I am unable to vend specific answers for nonspecific misunderstandings, misrepresentations or in the presence of material ambiguity. Though I am sympathetic toward that which I gather you might appreciate, I will be broadly unable to accommodate my understanding of the request now or in the future on technical pull requests afflicted by any the aforementioned.

Thanks for the clear listing. I'll address or correct each inline (leaving some inaccurate phrasing in place to focus on the more salient aspects):

> If an object has ~a ttl of 1000~ 1 second of ttl remaining and I call this function to set it to 2000 milliseconds:

* increase_to_milliseconds: The first call sets it to ~2000~ $(( `date -u +%s` + 2 )) and subsequent calls ~leave it at 2000~ set it to $(( `date -u +%s` + 2 )).
* decrease_to_milliseconds: The first call leaves it at 1000 as do subsequent calls
* overwrite_to_milliseconds: The first call sets it to ~2000~ $(( `date -u +%s` + 2 )) and subsequent calls ~leave it at 2000~ set it to $(( `date -u +%s` + 2 )).

> If an object has ~a ttl of 3000~ 3 seconds of ttl remaining and I set it to 2000:

* increase_to_milliseconds: The first call leaves the ttl at 3000 ~as do subsequent calls.~ and subsequent calls, if they come at least 1 second later, update the ttl to $(( `date -u +%s` + 2 )).
* decrease_to_milliseconds: The first call sets it to 2000 and subsequent calls leave it at 2000.
* overwrite_to_milliseconds: The first call sets it to 2000 and subsequent calls ~leave it at 2000.~ update the ttl to $(( `date -u +%s` + 2 )).

I hope these clarifications help you identify the mental model disconnect, but if there is any additional clarification I can provide feel free to reach out.

nand4011: I'll update the pr. The docs on the definition for that function don't specify whether it refreshes the ttl, and I assumed it did not. In the future, feel free to simply state why my assumption is incorrect.

kvcache: > The docs on the definition for that function don't specify whether it refreshes the ttl, and I assumed it did not.

I'm not sure what you mean by this and it might indicate a continued misunderstanding. In your examples you seem to (correctly) assume the ttl's in the request object represent relative ttl values. The dimension of time, however, does not pause between api invocations, so a relative duration value would be interpreted differently based on the context of different arrival times.

> In the future, feel free to simply state why my assumption is incorrect.

As I tried to clarify before, I will be unable to do this now or in the future _particularly_ when your assumptions are not made clear. This conversation is no longer relevant to this PR. We can follow up on slack if you're interested in pursuing it further.

nand4011:
While idempotent and retryable are a close match, they aren't an exact one. Since the original purpose for this was aiding retry logic I'm closing it.

PR #152, opened 2023-04-05
feat: custom message extensions

kvcache:
For dynamic inspection of messages, the inaugural extension is just
a retry semantic. The quality of retryability is  a little different
from idempotence. Each message may choose if we think it should be
automatically retried independently of whether or not it pedantically
matches a mathematical definition.

### DIFF
--- a/proto/extensions.proto
+++ b/proto/extensions.proto
@@ -1,0 +1,20 @@
+syntax = "proto3";
+
+import "google/protobuf/descriptor.proto";
+
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_package = "grpc.extensions";
+option csharp_namespace = "Momento.Protos.Extensions";
+
+// A hint so you can decide a little more in the abstract "can this be retried?""
+enum RetrySemantic {
+    // Never retry this message without telling the user. (you should infer this as the default)
+    NotRetryable = 0;
+    // You can retry this without surfacing an error to the user.
+    Retryable = 1;
+}
+
+extend google.protobuf.MessageOptions {
+    // Can this message be re-driven without an error?
+    RetrySemantic retry_semantic = 50000;
+}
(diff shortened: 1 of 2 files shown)


proto/extensions.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
```
bruuuuuuuce, line 1: do we also need to publish this file like the rest of the proto files?

https://github.com/momentohq/client-protos/blob/main/javascript/generate_protos.sh#L5

proto/extensions.proto
```diff
@@ -0,0 +1,20 @@
...
+enum RetrySemantic {
+    // Never retry this message without telling the user. (you should infer this as the default)
+    NotRetryable = 0;
+    // You can retry this without surfacing an error to the user.
+    Retryable = 1;
+}
+
+extend google.protobuf.MessageOptions {
+    // Can this message be re-driven without an error?
+    RetrySemantic retry_semantic = 50000;
```
bruuuuuuuce, line 19: does the `50000` have any meaning? Or is it just that high because we are extending the `MessageOptions` and don't want to interfere with any messages they are sending within it?

kvcache: It is the latter. You'll find examples on the internet starting around this value, sometimes higher, and it seemed no less reasonable than any other implausibly high number.

proto/extensions.proto
kvcache: yeah, I expected to encounter more of these lists to bite me tbh.

kvcache: I hoped that the include path would address it but :-/ oh well

kvcache: https://github.com/momentohq/client-protos/actions/runs/4623819429/jobs/8178018886?pr=152#step:7:31

nope, this is an issue with the descriptor proto generated code.

PR #154, opened 2023-04-18
feat: Add protos for RefreshApiToken Api

rlinehan:
This API is used the "refresh" - i.e. extend the expiration length - of an API token. It takes in an existing `api_key` and `refresh_token` and returns back a new api key and refresh token. The response for this is the same as for the `GenerateApiToken` API.


kvcache (approved):
looks good to me

ben-kugler (approved):

PR #156, opened 2023-05-03
chore: update Javascript gRPC/protobuf dependencies

pgautier404:


javascript/package-lock.json
```diff
@@ -20,9 +20,9 @@
       }
     },
     "node_modules/@grpc/grpc-js": {
-      "version": "1.7.3",
-      "resolved": "https://registry.npmjs.org/@grpc/grpc-js/-/grpc-js-1.7.3.tgz",
-      "integrity": "sha512-H9l79u4kJ2PVSxUNA08HMYAnUBLj9v6KjYQ7SQ71hOZcEXhShE/y5iQCesP8+6/Ik/7i2O0a10bPquIcYfufog==",
+      "version": "1.8.14",
+      "resolved": "https://momento-prod-401011790710.d.codeartifact.us-west-2.amazonaws.com/npm/npm-upstream/@grpc/grpc-js/-/grpc-js-1.8.14.tgz",
```
cprice404, line 24: Did you do a `mm dev login` that logged you in to our codeartifact repo?

You need to stash the changes to this file and remove your ~/.npmrc file and run npm install again so that all of the resolved urls in here point to registry.npmjs.org.

cc: @rishtigupta this is the thing I was telling you about yesterday - builds will fail on github if we accidentally end up with references to our codeartifact repo in our public/OSS projects.

cprice404 (changes requested):
bad npm artifact urls in your package-lock.json file, see my comment in that file

javascript/package.json
```diff
@@ -16,13 +16,13 @@
   "devDependencies": {
     "@tsconfig/node16": "1.0.2",
     "@types/node": "16.10.3",
-    "protoc-gen-ts": "0.6.0",
-    "typescript": "4.4.3"
+    "protoc-gen-ts": "0.8.6",
```
krispraws, line 19: Did you confirm that this version is ok since 0.8.5 was a breaking change?

krispraws (approved):

javascript/package.json
cprice404: The easiest way to confirm that is going to be to release this artifact and then try to upgrade the JS SDKs to consume it.  If it doesn't work then they will skip this version, and we will revert this if necessary.

cprice404 (approved):

PR #157, opened 2023-05-09
fix: remove ts files from bundled code

bruuuuuuuce:
Previously we had been bundling .ts files with our protos. This was causing applications consuming the protos to try and recompile/lint check the bundled ts files. This pr fixes this by generating the protos into a `src` folder, and then running `tsc` to bundle this `src` folder into the `dist` folder which will get published.

TESTING

Checked out this commit https://github.com/momentohq/client-sdk-nodejs/pull/472/commits/b5115d59936aa58e45aae53fc797ff34907b61dc which was facing ts build issues. In this branch ran

- npm run build
- npm pack

and then in the nodejs sdk commit above, installed the packed version and rebuilt it. No more ts errors


kvcache (commented):
I think this looks right. Thanks Matt.

Do .d.ts type "safety" and editor documentation pop ups still happen on leaf projects even after clearing local package caches? If it's all looking good I'm +1 on this but you'll probably want @cprice404 to approve.

cprice404 (commented):
this looks right to me, thanks @bruuuuuuuce .  Do we need to do this for the web one too though?

bruuuuuuuce:
> this looks right to me, thanks @bruuuuuuuce . Do we need to do this for the web one too though?

Yep I believe so. I can add that to this pr or do a fast follow. Either works for me. I justed wanted to unblock @ben-kugler first

cprice404:
separate PR is fine then

PR #158, opened 2023-05-09
fix: remove ts files from bundled code for js-web

bruuuuuuuce:
This change is the same as https://github.com/momentohq/client-protos/pull/157, but for the grpc-web sdk. We no longer are bundling the `.ts` files in the package, and just bundling the `.d.ts` and `.js` files


cprice404 (approved):
:ty: :hero:

PR #159, opened 2023-05-10
feat: fix packaging for new grpc/protobuf versions

cprice404:
This commit fixes some packaging issues with the new versions
of grpc/protobuf and the recent changes to exclude the .ts files
from the dist.

In particular:

* Don't nest the `src` dir in the distributable
* A bunch of .d.ts files and .js files were missing from the web
  sdk


bruuuuuuuce (approved):
:shipit:

PR #163, opened 2023-05-16
feat: add permissions model to the auth token request proto

bruuuuuuuce:
This pr allows us to assign specific permissions to an auth token. v1 api tokens created without this `permissions` claim will be treated as `SuperUser` tokens with full permissions. Eventually, we will probably add additional restrictions to support more granular permissions

```
permissions: {
  cache: {
    role: readwrite

    [future stuff]
    key: base64?,
    key_utf8: string?,
    key_hash: sha1?,
    apis: ...
  },
  topic: {
    role: readwrite,

    [future stuff]
    topic: "if this is here it's limited to this topic"
  }
}
```

We will also add more roles in the future. MVP is to get a token that has data plane read/write access, so it is currently the only role we need. More details can be found in the notion doc https://www.notion.so/momentohq/Authz-67866217f45a4f3dbaf4c89cade335e9

### DIFF
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -70,10 +70,17 @@ message _GenerateApiTokenRequest {
     }
 
     // Aliases for categories of functionality.
-    enum Role {
-      PermitNone = 0;
-      // Restricts access to apis that read and write data: No higher level resource description or modification.
-      ReadWrite = 1;
+    enum CacheRole {
+      CachePermitNone = 0;
+      // Restricts access to apis that read and write data from caches: No higher level resource description or modification.
+      CacheReadWrite = 1;
+    }
+
+    // Aliases for categories of functionality.
+    enum TopicRole {
+      TopicPermitNone = 0;
+      // Restricts access to apis that read and write data from topics: No higher level resource description or modification.
+      TopicReadWrite = 1;
     }
 
     string auth_token = 3;
@@ -82,14 +89,14 @@ message _GenerateApiTokenRequest {
     // on the request, then the token is assumed to have "SuperUser" permissions
     message Restrictions {
       message CacheRestrictions {
-        Role role = 1;
+        CacheRole role = 1;
         // eventually could scope this down to a specific cache name or key
         // string cache = "an example cache";
         // string caches = ["cache1", "cache2"]
       }
 
       message TopicsRestrictions {
-        Role role = 1;
+        TopicRole role = 1;
         // eventually could scope this down to a specific topic name
         // string topic = "an example topic";
         // string topics = ["topic1", "topic2"]
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -99,7 +99,7 @@ message _GenerateApiTokenRequest {
       TopicsRestrictions topic = 2;
     }
 
-    Permissions permissions = 4;
+    repeated Permissions permissions = 4;
 }
 
 message _GenerateApiTokenResponse {
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -95,8 +95,10 @@ message _GenerateApiTokenRequest {
         // string topics = ["topic1", "topic2"]
       }
 
-      CacheRestrictions cache = 1;
-      TopicsRestrictions topic = 2;
+      oneof kind {
+          CacheRestrictions cache = 1;
+          TopicsRestrictions topic = 2;
+      }
     }
 
     repeated Restrictions restrictions = 4;
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -78,9 +78,9 @@ message _GenerateApiTokenRequest {
 
     string auth_token = 3;
 
-    // Contains the permissions that the token will have for each of our services. If it doesn't exist
+    // Contains the restrictions that the token will have for each of our services. If it doesn't exist
     // on the request, then the token is assumed to have "SuperUser" permissions
-    message Permissions {
+    message Restrictions {
       message CacheRestrictions {
         Role role = 1;
         // eventually could scope this down to a specific cache name or key
@@ -99,7 +99,7 @@ message _GenerateApiTokenRequest {
       TopicsRestrictions topic = 2;
     }
 
-    repeated Permissions permissions = 4;
+    repeated Restrictions restrictions = 4;
 }
 
 message _GenerateApiTokenResponse {
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -69,8 +69,11 @@ message _GenerateApiTokenRequest {
         Expires expires = 2;
     }
 
+    // Aliases for categories of functionality.
     enum Role {
-      ReadWrite = 0;
+      PermitNone = 0;
+      // Restricts access to apis that read and write data: No higher level resource description or modification.
+      ReadWrite = 1;
     }
 
     string auth_token = 3;
@@ -78,22 +81,22 @@ message _GenerateApiTokenRequest {
     // Contains the permissions that the token will have for each of our services. If it doesn't exist
     // on the request, then the token is assumed to have "SuperUser" permissions
     message Permissions {
-      message CachePermissions {
+      message CacheRestrictions {
         Role role = 1;
         // eventually could scope this down to a specific cache name or key
         // string cache = "an example cache";
         // string caches = ["cache1", "cache2"]
       }
 
-      message TopicsPermissions {
+      message TopicsRestrictions {
         Role role = 1;
         // eventually could scope this down to a specific topic name
         // string topic = "an example topic";
         // string topics = ["topic1", "topic2"]
       }
 
-      CachePermissions cache = 1;
-      TopicsPermissions topic = 2;
+      CacheRestrictions cache = 1;
+      TopicsRestrictions topic = 2;
     }
 
     Permissions permissions = 4;
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -69,8 +69,34 @@ message _GenerateApiTokenRequest {
         Expires expires = 2;
     }
 
-    // session token recieved from `momento login` command
-    string session_token = 3;
+    enum Role {
+      ReadWrite = 0;
+    }
+
+    string auth_token = 3;
+
+    // Contains the permissions that the token will have for each of our services. If it doesn't exist
+    // on the request, then the token is assumed to have "SuperUser" permissions
+    message Permissions {
+      message CachePermissions {
+        Role role = 1;
+        // eventually could scope this down to a specific cache name or key
+        // string cache = "an example cache";
+        // string caches = ["cache1", "cache2"]
+      }
+
+      message TopicsPermissions {
+        Role role = 1;
+        // eventually could scope this down to a specific topic name
+        // string topic = "an example topic";
+        // string topics = ["topic1", "topic2"]
+      }
+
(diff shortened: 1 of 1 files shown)


proto/auth.proto
```diff
@@ -69,8 +69,34 @@ message _GenerateApiTokenRequest {
...
+    enum Role {
+      ReadWrite = 0;
+    }
+
+    string auth_token = 3;
+
+    // Contains the permissions that the token will have for each of our services. If it doesn't exist
+    // on the request, then the token is assumed to have "SuperUser" permissions
+    message Permissions {
+      message CachePermissions {
```
bruuuuuuuce, line 81: maybe this should be `CacheRestrictions`?

proto/auth.proto
```diff
@@ -69,8 +69,34 @@ message _GenerateApiTokenRequest {
         Expires expires = 2;
     }
 
-    // session token recieved from `momento login` command
-    string session_token = 3;
+    enum Role {
+      ReadWrite = 0;
+    }
```
kvcache, line 74:
```suggestion
    // Aliases for categories of functionality.
    enum Role {
      // Restricts access to apis that read and write data: No higher level resource description or modification.
      ReadWrite = 0;
    }
```

proto/auth.proto
kvcache: imo that would be the thing to do.

proto/auth.proto
```diff
@@ -69,8 +69,34 @@ message _GenerateApiTokenRequest {
         Expires expires = 2;
     }
 
-    // session token recieved from `momento login` command
-    string session_token = 3;
+    enum Role {
+      ReadWrite = 0;
```
kvcache, line 73: 🤔 this is going to be a default value deserialized for lots of stuff. Since this goes in the restrictions block I think you want a `PermitNone = 0` and a `ReadWrite = 1`?

bruuuuuuuce: ah yea good point

proto/auth.proto
```diff
@@ -69,8 +69,37 @@ message _GenerateApiTokenRequest {
         Expires expires = 2;
     }
 
-    // session token recieved from `momento login` command
-    string session_token = 3;
+    // Aliases for categories of functionality.
+    enum Role {
```
mjindalmotiv8, line 73: I thought we did not want to use the word Role and wanted to use scope?

proto/auth.proto
rlinehan: if we want to go with the language `restrictions` ok, but personally I find it breaks my brain a little? Using `permissions` here is more what I'm used to.

proto/auth.proto
rlinehan: I don't know that the wording really matters tbh

mjindalmotiv8: It mattered to @kvcache yesterday so want to make sure he is ok here

proto/auth.proto
```diff
@@ -69,8 +69,37 @@ message _GenerateApiTokenRequest {
...
+    message Permissions {
+      message CacheRestrictions {
+        Role role = 1;
+        // eventually could scope this down to a specific cache name or key
+        // string cache = "an example cache";
+        // string caches = ["cache1", "cache2"]
+      }
+
+      message TopicsRestrictions {
+        Role role = 1;
```
rlinehan, line 92: A protobuf-type question: if this role enum is the same role enum for caches, can we later on break it apart into separate enums if we need to?

Like say later on I want to add a `foo` role for Caches but not for Topics - could we do that?

proto/auth.proto
```diff
@@ -69,8 +69,37 @@ message _GenerateApiTokenRequest {
...
+      PermitNone = 0;
+      // Restricts access to apis that read and write data: No higher level resource description or modification.
+      ReadWrite = 1;
+    }
+
+    string auth_token = 3;
+
+    // Contains the permissions that the token will have for each of our services. If it doesn't exist
+    // on the request, then the token is assumed to have "SuperUser" permissions
+    message Permissions {
```
danielamiao, line 83: We should probably just name this `Restrictions` since the inner content are all specific types of restrictions not actual permissions (allowed things).

proto/auth.proto
danielamiao: `Action` is another option

kvcache: role meant something different in the context of yesterday's conversation. I think in this narrow position it doesn't mean something terribly different from "role" but maybe future kenny will disagree.

kvcache: I think it'sfine 🤷

proto/auth.proto
```diff
@@ -69,8 +69,37 @@ message _GenerateApiTokenRequest {
         Expires expires = 2;
     }
 
-    // session token recieved from `momento login` command
-    string session_token = 3;
+    // Aliases for categories of functionality.
+    enum Role {
+      PermitNone = 0;
```
rlinehan, line 74: I'm unsure of the necessity for `permit none`

```
{
    permissions: {
      cache: {
          role: read
      }
}
```

To me means only cache read is allowed; no topics or control plane is allowed. I don't think you should have to do
```
{
   permissions: {
      cache: {
         role: read
       }
     topics: {
        role: permit none
     }
     control: {
        role: permit none
    }
  }
}
```

Because if we do this, then say we add a new family of apis called idk "orgs." When I originally created my token, my intention was to only have it be able to read from caches and do nothing else. But since `orgs` isn't in this list with a `permit none` then the token would be allowed to access these new APIs.

danielamiao (approved):

proto/auth.proto
```diff
@@ -69,8 +69,37 @@ message _GenerateApiTokenRequest {
...
+        // eventually could scope this down to a specific topic name
+        // string topic = "an example topic";
+        // string topics = ["topic1", "topic2"]
+      }
+
+      CacheRestrictions cache = 1;
+      TopicsRestrictions topic = 2;
+    }
+
+    repeated Restrictions restrictions = 4;
```
kvcache, line 102:
```suggestion
      oneof kind {
          CacheRestrictions cache = 1;
          TopicsRestrictions topic = 2;
      }
    }

    repeated Restrictions restrictions = 4;
```

since this is repeated now, this will be more flexible and should be significantly easier to parse (existence test won't be weird) without making it any harder to produce it.

bruuuuuuuce: ah yea thats true, good call

proto/auth.proto
bruuuuuuuce: I think the concern is that by default the enum parses to 0, so if we unmarshal it and it doesnt exist, then it would be set to `ReadWrite`. But if its unset is should be `NoPerms`

rlinehan: oh right protobuf defaults. ok, I think that makes sense!

proto/auth.proto
bruuuuuuuce: no you would need to add another `TopicRole role2 = 2`, and deprecate the first role

proto/auth.proto
danielamiao: oh man, this is a good point, if the default behavior is `permit none` shouldn't everything be named `permissions`?

`permissions`
`cachePermissions`
`topicPermissions`

the absence of a permission means deny

proto/auth.proto
kvcache: Yes - you can read the proto3 specification, but the short answer is this looks like `$tag_byte$enum_byte` on the wire. So it would be a series `12` for topicrestrictions readwrite. You could flip it to another enum, but that enum would need ordinals matching every previously possible value to be set. So we should never shoehorn in something that does not fit :-)

proto/auth.proto
bruuuuuuuce: its a whitelist, not a blacklist, which is where the word `restriction` is coming from

proto/auth.proto
kvcache: matt's way is safer, and also documented in proto3 upgrade patterns

rlinehan: ok I guess my question is should these be the same enum, or should we just have them be separate enums from the start?

proto/auth.proto
rlinehan: this is why naming things "restrictions" confuses the hell out of me 😬

proto/auth.proto
bruuuuuuuce: having them be separate enums would make it more flexible i think

proto/auth.proto
kvcache: if `permissions` is nonzero length, then the token is restricted.

The first layer is "permit" semantic, allowing access to cache/topic/whatever feature set.

The second layer is "restrict" semantic, narrowing access to the cache/topic/whatever feature set.

proto/auth.proto
kvcache: sure, is fine

proto/auth.proto
danielamiao: lol both work, it's a yanny vs laurel situation:

- `restrictions` : this is the set of things users are "restricted to do", everything else is not allowed
- `permissions`: this is the set of things users are "permitted to do", everything else is not allowed

I don't care which one, I just think we should be consistent and not use both

proto/auth.proto
```diff
@@ -69,8 +69,46 @@ message _GenerateApiTokenRequest {
...
+    // Aliases for categories of functionality.
+    enum CacheRole {
+      CachePermitNone = 0;
+      // Restricts access to apis that read and write data from caches: No higher level resource description or modification.
+      CacheReadWrite = 1;
+    }
+
+    // Aliases for categories of functionality.
+    enum TopicRole {
+      TopicPermitNone = 0;
```
bruuuuuuuce, line 81: thought this was interesting

Screen Shot 2023-05-16 at 4 29 46 PM

rlinehan: Oh good catch!

proto/auth.proto
```diff
@@ -69,8 +69,46 @@ message _GenerateApiTokenRequest {
...
+      // Restricts access to apis that read and write data from topics: No higher level resource description or modification.
+      TopicReadWrite = 1;
+    }
+
+    string auth_token = 3;
```
rlinehan, line 86: This is totally a nitpick... could you move this up above the enums? Otherwise it's just sort of stranded here having nothing to do with permissions 😆 

Separately... how are we going to role out this name change on the server? I think since the console is already using this api we maybe need to deprecate `session_token` and add this as a new value?

rlinehan (commented):
I don't love the "restrictions" naming but I also don't care that much.

I do however have a concern about renaming `session_token` to `auth_token` safely.....

Otherwise, I'm good on this 👍 .

proto/auth.proto
cprice404: as long as the type and the index are the same, the name doesn't matter... protobuf just packs the values as binary data, names aren't used in the serialization

rlinehan: oh, ok, cool!

bruuuuuuuce:
> I do however have a concern about renaming `session_token` to `auth_token` safely.....

renaming the proto field names are safe changes. Changing the typing is the unsafe proto changes.

https://stackoverflow.com/questions/45431685/protocol-buffer-does-changing-field-name-break-the-message

rlinehan (approved):

PR #165, opened 2023-05-17
fix: use permissions model instead of restrictions

bruuuuuuuce:


proto/auth.proto
```diff
@@ -85,30 +85,37 @@ message _GenerateApiTokenRequest {
...
+        CacheRole role = 1;
+      }
+
+      message TopicPermissions {
+        TopicRole role = 1;
       }
     }
 
-    repeated Restrictions restrictions = 4;
+    repeated Permissions permissions = 4;
```
cprice404, line 118: i think this is no longer repeated now?

rlinehan (approved):

proto/auth.proto
bruuuuuuuce: yes youre right

bruuuuuuuce, line 118:
```suggestion
    Permissions permissions = 4;
```

rlinehan (approved):

cprice404 (approved):

PR #166, opened 2023-05-22
fix: python - apply relative import substitution for all generated gRPC service code

malandis:
The python generated code has a shortcoming where the gRPC service
modules import the protobuf types using absolute imports. For us to package
our wire types as an installable package, we change those absolute imports
to relative imports in a post-processing step.

The previous version patched the imports on the cache and control
client gRPC code using a hard-coded list of modules. Because we have since
added more gRPC services (auth), we would have had to add that
replacement to the list; however, we forgot to do this.

Because remembering to update this list is error prone, we change the
code to instead perform this replacement in *any* gRPC service
generated code. This was the original intended behavior.


cprice404 (approved):
🙈

PR #174, opened 2023-06-15
chore: update rust, remove protoc user dependency

kvcache:
* Move proto generation into a separate, unpublished, non-dependent sub-package.
* Move momento-protos into explicit sub-package.
* Create workspace.
* Update PR checks and release publish invocations.
* Update Tonic dependencies.
* Remove tonic-build build-time dependency from users of momento-protos.
* Remove protoc build-time dependency from users of momento-protos.

I checked this locally with another project with a dependency on `momento-protos`.
The import paths are unchanged, and the version bump has the intended effect.


PDXKimani (dismissed):

rust/momento-protos/src/_.rs
```diff
@@ -0,0 +1,29 @@
+/// A hint so you can decide a little more in the abstract "can this be retried?""
```
cprice404, line 1: so the stuff in this dir is the generated code, yes? if so, does that mean a dev needs to do something to update this whenever they update the actual proto files? will they know that and will they know what they need to do?

maybe slightly orthogonal question, does this dir need to be committed to vcs?

kvcache: Yes. Yes. Yes. Yes. :-)

When a developer opens a pr with a proto update and without having done `cargo build` on the workspace, the check will fail. When they look at the check that failed they will see a comment telling them about the 2 most likely ways they triggered the failure.

That this directory is checked in is a primary goal of this PR. Without this directory, our users have a build-time dependency on installing protoc on their own build hosts purely on our account.

cprice404: couldn't the CI process do the cargo build and then publish the generated artifacts to crates.io? wouldn't that be the best of both worlds?

cprice404: like, i get why you are trying to remove the deps, and that seems good and noble, but i don't really understand why that would necessitate that we check the generated code into vcs.

kvcache: nondeterminism, opacity and independent variance between dev environments and the builder seems less good than clear intentionality enforced by automation. It is my opinion that the best of both worlds here is to check in the generated source, since we're generating it separately, and enforce that the generated source matches the checked in source. It's nice to be able to look at a package's source code. What's the concern here?

cprice404: it's just weird that people need to go do a cargo build after they change the protos. we don't require this for any of the other languages.

if the CI job did the cargo build and then published the resulting, generated code to crates.io, consumers of the package would still have access to the source code, would they not? this is how it works for all the other languages

this just seems unusual and different compared to all the other languages as well as the ways that I have always handled these kinds of things in the past. checking in generated source is kind of a weird and anomalous thing to do and in this case it seems like there's a super easy way to avoid it, with no discernable downside.

but you seem to feel strongly about this and i don't really want to debate it so it's fine if you want to merge it

kvcache: I do feel strongly that checking in the generated files is better, but you seem to feel strongly it isn't!

I don't feel strongly enough to try to make it happen. I'll keep the dependency management benefits but remove the checked in generated file. See how the update grabs you.

cprice404 (approved):
dunno why the kotlin build is failing, do you?

kvcache:
> @cprice404 dunno why the kotlin build is failing, do you?

Vaguely, yes. From slack:
> I was not able to open this PR via a branch on momentohq. I was given an odd error message about the fork already existing - I'm not sure if it's some permissions thing or what. So the pr checks are going to fail due to missing credentials on account of public fork secret access.

PR #179, opened 2023-06-22
feat: proto changes for fine grained authorization

krispraws:
- Adds ReadOnly role types for Cache and Topic permissions
- Permissions can specify all caches/topics (`Any`) or a specific cache/topic name. This can be extended in future to support partial matches.
- I had to change some scripts because the existing ones did not work on my Mac.

### DIFF
--- a/proto/auth.proto
+++ b/proto/auth.proto
@@ -74,6 +74,8 @@ message _GenerateApiTokenRequest {
       CachePermitNone = 0;
       // Restricts access to apis that read and write data from caches: No higher level resource description or modification.
       CacheReadWrite = 1;
+      // Restricts access to apis that read from caches: No higher level resource description or modification.
+      CacheReadOnly = 2;
     }
 
     // Aliases for categories of functionality.
@@ -81,6 +83,8 @@ message _GenerateApiTokenRequest {
       TopicPermitNone = 0;
       // Restricts access to apis that read and write data from topics: No higher level resource description or modification.
       TopicReadWrite = 1;
+      // Restricts access to apis that read from topics: No higher level resource description or modification.
+      TopicReadOnly = 2;
     }
 
     string auth_token = 3;
@@ -106,12 +110,31 @@ message _GenerateApiTokenRequest {
         TopicPermissions topic_permissions = 2;
       }
 
+      message Any {}
+
+      message CacheResource {
+        oneof kind {
+          Any any = 1;
+          string cache_name = 2;
+        }
+      }
+
       message CachePermissions {
         CacheRole role = 1;
+        CacheResource cache = 2;
+      }
+
+      message TopicResource {
+        oneof kind {
+          Any any = 1;
+          string topic_name = 2;
+        }
       }
 
       message TopicPermissions {
         TopicRole role = 1;
+        CacheResource cache = 2;
+        TopicResource topic = 3;
       }
     }
 
--- a/install_protoc.sh
+++ b/install_protoc.sh
@@ -4,17 +4,22 @@ set -x
 
 VERSION=3.18.1
 
-sudo apt-get -y install zip tree
-curl -L https://github.com/protocolbuffers/protobuf/releases/download/v$VERSION/protoc-$VERSION-linux-x86_64.zip -o protoc.zip
-unzip -o protoc.zip -d protoc
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  # Install protoc for local development on MacOS (requires Homebrew)
+  brew install zip tree
+  PROTOC_ZIP=protoc-$VERSION-osx-x86_64.zip
+else # Linux
+  sudo apt-get -y install zip tree
+  PROTOC_ZIP=protoc-$VERSION-linux-x86_64.zip
+fi
 
-mkdir temp
-pushd temp
-  curl -L https://github.com/protocolbuffers/protobuf/releases/download/v$VERSION/protoc-$VERSION-linux-x86_64.zip -o protoc.zip
-  unzip -o protoc.zip -d protoc
-  sudo mv protoc/bin/* /usr/local/bin/
-  sudo mv protoc/include/* /usr/local/include/
-popd
-rm -rf temp
+curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v$VERSION/$PROTOC_ZIP
+sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc
+sudo chmod a+rx /usr/local/bin/protoc
+sudo unzip -o $PROTOC_ZIP -d /usr/local 'include/*'
+sudo chmod -R a+rx /usr/local/include/google
+rm $PROTOC_ZIP
 
 tree /usr/local/include/google
+echo "Installed protoc $VERSION"
+protoc --version
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -1,6 +1,5 @@
 # Generating definitions
 ```bash
 npm install
-mkdir dist
-PATH=node_modules/protoc-gen-ts/bin/:$PATH protoc -I=../proto --ts_out=dist cacheclient.proto controlclient.proto cachepubsub.proto
+./generate_protos.sh
 ```
--- a/javascript/generate_protos.sh
+++ b/javascript/generate_protos.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 set -e
 set -x
-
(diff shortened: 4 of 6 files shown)


malandis (dismissed):
Very cool!

proto/auth.proto
```diff
@@ -106,12 +110,31 @@ message _GenerateApiTokenRequest {
         TopicPermissions topic_permissions = 2;
       }
 
+      message Any {}
+
+      message CacheResource {
+        oneof kind {
+          Any any = 1;
+          string cache_name = 2;
```
krispraws, line 118: @danielamiao , @kvcache - I chose to only allow one cache name per permission for simplicity and have multiple permissions if the user wants to add more caches, but I have been wondering if this will blow up our token sizes and if it would be better to allow a list of cache names per Permission. Same applies for topics though in that case mapping between list of cache names and list of topic names gets tricky. But we can enforce in the service that topic permissions can only specify Any or a single cache name. Thoughts?

cprice404: i'm a slight +1 on making it an array but I don't feel strongly

cprice404 (approved):
lgtm pending resolution of the question about possibly supporting multiple caches/topics in one rule.

proto/auth.proto
krispraws: I can see how it would be nicer for the user, but my main concern is that a TopicPermission like below is confusing. Does it mean it applies to the cross product of the two lists, so all of these topics are valid? : (abc, xyz) (abc, uvw), (def, xyz) and so on
```
topicRole: ReadOnly,
cache : { cache_names: ["abc", "def", "ghi"] },
topic: { topic_names: ["xyz", "uvw"]}
```

ben-kugler: I'm a +1 on __not__ having this be an array, the more simple and clear these permissions are, the better. As Ramya also mentioned, token sizing is worrying and making some random arbitrary limit on number of topic/cache names doesn't seem great. 

But nonetheless, this is something we could revisit later and add if need be.

ben-kugler (dismissed):

proto/auth.proto
cprice404: how does making it "not an array" solve the cross-product issue?

in that model wouldn't you have perms like this:

```
[cache: "abc"], [cache: "def"], [topic: "xyz"], [topic: "uvw"]
```

wouldn't the user still have the same question in that case?

hopefully i'm just missing something, and my apologies if so!

krispraws: It solves it because TopicPermission contains both cache and topic and they can only be Any | scalar name. e.g.
```
[cache: "abc"], [cache: "def"], [cache: "abc", topic: "xyz"], [cache: Any, topic: "uvw"]
```

ben-kugler (dismissed):

proto/auth.proto
danielamiao: Hm, what does `[cache: Any, topic: "uvw"]` mean? Does it mean any topics named `uvw` under all caches? I thought that wasn't possible/allowed?

krispraws: It is technically feasible - @ben-kugler please keep me honest. We have a [scenario](https://www.notion.so/momentohq/Authorization-Scenarios-f1cea5c1caee4523b3ebe85687c65a55?pvs=4#a556cd2953d44700a2260d7fd2f0941c) for All Topics on All Caches `[cache: Any, topic: Any]` , and I felt that  Topic X on All Caches `[cache: Any, topic: "uvw"] ` was just a special case of it.  We can add validation server side to disallow it if needed.

ben-kugler: Yea, seems very feasible, we would only restrict on the topic permission. Doesn't seem inherently harmful, the sdk's will make it clear so customers aren't surprised by the behavior. I don't see any reason not to allow it.

cprice404: so then couldn't `caches` be an array, but the `cache` field inside of a TopicPermission could still be singular, and `topics` field inside of a TopicPermission could be an array?

I'm 100% on board with making sure it's simple for the user to understand but just haven't yet wrapped my head around whether it's impossible to achieve that while also allowing them to specify more than one cache/topic in a perm :)

krispraws: it is not impossible - it just means I'd need to model the cache identifier within topic permission differently than the cache identifier for a cache permission. I wanted to re-use the same structure because it would be simpler to extend it in future if we want to add prefixes or regexes, but using 2 structures is definitely possible.

cprice404: if you pulled the `Any` up a level, then you could re-use the structure, right? (in fact if you want to be able to model something more complex than just names, won't you need for `Any` to be at a different level in the object hierarchy than where it currently is? where it is currently at the same level as `cache_name`?)

I don't feel super strongly about this so please feel free to just make a call if I am distracting/slowing you down :)

krispraws: @cprice404 - is this what you are suggesting?
```
    message PermissionsType {
      oneof kind {
        CachePermissions cache_permissions = 1;
        TopicPermissions topic_permissions = 2;
      }

      message All {}

      message CacheNames {
        repeated string names = 1;
      }

      message CachePermissions {
        CacheRole role = 1;
        oneof caches {
          All all = 2;
          CacheNames cache_names = 3;
        }
      }

      message TopicNames {
        repeated string names = 1;
      }

      message TopicPermissions {
        TopicRole role = 1;
        oneof cache {
          All all_caches = 2;
          string cache_name = 3;
        }
        oneof topics {
          All all_topics = 4;
          TopicNames topic_names = 5;
        }

      }
    }
```

cprice404: So with your current code you could just change these two:
```
      message CachePermissions {
        CacheRole role = 1;
        repeated CacheResource cache = 2;
      }
      message TopicPermissions {
        TopicRole role = 1;
        CacheResource cache = 2;
        repeated TopicResource topic = 3;
      }
    }
```

Then you are supporting multiple caches for a Cache Permissions, multiple topics (but a single cache) for a Topic Permission, and still achieving your goal of re-using `CacheResource`?

My half-baked comments about moving `Any` up were just from thinking about whether it would be weird for people to be able to put two `Any`s in a `repeated`, but after further thought I don't think that's a big deal, they could technically do the same thing without the `repeated` by just adding two instances of `CachePermissions` with the cache set to `Any`.

I had been thinking of something along the lines of:

```
      message CacheResource {
        oneof kind {
          Any any = 1;
          CacheSelector cache = 2;
        }
      }

     message CacheSelector {
        one of kind {
             string cache_name = 1;
             string cache_prefix = 2;
             string regex = 3;
        }
     }
```

to separate the Any from the actual selectors. But after having typed it out I don't think it makes sense :)

I do think the `repeated` is worth considering but again I don't feel strongly about it!

krispraws: I like your idea. Thank you! I'll make that first change. re: the CacheSelector, I think we can add that in future like
```
message CacheResource {
        oneof kind {
          Any any = 1;
          string cache_name = 2; 
          CacheSelector selector = 3;
        }
      }
```

cprice404: and yes, what you just posted is roughly where I think my brain was heading with my comment about pulling the `All`/`Any` up a level but I'm not sure it's necessary now that i've looked at the options all typed out :)

krispraws: @cprice404 & others, after discussing with Ruth and Ben, I decided to start simple with a single cache and topic name per permission. I moved All up one level and renamed CacheResource to CacheSelector (thanks to Chris for that name). If we want to support multiple cache names or a cache prefix in future, we will extend CacheSelector as 
```
message CacheNames { repeated string names = 1; }
message CacheSelector {
        oneof kind {
          string cache_name = 1;
          CacheNames cache_names = 2;
          string cache_prefix = 3;
        }
      }
```
And if/when we add that, we can either decide to allow multiple cache names in a single TopicPermission or restrict it in code.

rlinehan (dismissed):

proto/auth.proto
cprice404: this sounds fine to me, as long as we make sure to model the SDK code in a forward compatible way. I think you are doing this in your current JS SDK PR via a union type where we can introduce `| CacheNames` in the future. if so, wfm

ben-kugler (dismissed):

proto/auth.proto
krispraws: JS changes for reference: https://github.com/momentohq/client-sdk-javascript/pull/598/files#diff-f3d3e3f16b697af5c41a0000ea847dc7f268803939242f9f9c7efb1a5b87ca02
SDK code will be forward compatible

rlinehan (approved):

PR #181, opened 2023-06-30
feat: bump grpc js to version 1.8.17

bruuuuuuuce:
Features/fixes that come with this upgrade

- Ensure status and error events are consistently emitted asynchronously 
- Fix a memory leak that could result from a specific pattern of recursive function calls
- Fix missing transport trace logs
- Disallow pick_first LB policy as the direct child of an outlier_detection LB policy

### DIFF
--- a/javascript/package-lock.json
+++ b/javascript/package-lock.json
@@ -1,14 +1,15 @@
 {
   "name": "@gomomento/generated-types",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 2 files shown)


cprice404:
in your commit message: are those notes from the upstream changelog? if so, thanks very much for including them, that is the right thing to do. But I would add a sentence to your description/commit message before those bullet points that clearly states that those changes are coming in from the upstream library.


Also, we can't do a version change of a major underlying component like this in a .z release, this needs to be `feat` so that it will be a `.y` release.

cprice404 (approved):

PR #183, opened 2023-08-03
fix: Remove self-generated google common protos

nand4011:
Stop generating google common protos. A customer is getting conflicts because they are colliding with those in the proto-google-common-protos dependency.

Change java artifact developer to the generic momento dev to match the java SDK.

Temporarily disable conditional releases of the java protos and add debug statements because java appears to never be released.

### DIFF
--- a/java/build.gradle.kts
+++ b/java/build.gradle.kts
@@ -26,12 +26,14 @@ dependencies {
     implementation("io.grpc:grpc-api")
     implementation("io.grpc:grpc-protobuf")
     implementation("com.google.protobuf:protobuf-java")
+    implementation("com.google.api.grpc:proto-google-common-protos:2.9.0") // version pulled from protobuf-java
     implementation("com.google.guava:guava:31.1-android") // version pulled from protobuf-java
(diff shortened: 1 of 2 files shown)


nand4011:
I was able to run the integration tests in the Java SDK using this as a local dependency. I'll do more testing once it's released.

java/build.gradle.kts
```diff
@@ -26,12 +26,14 @@ dependencies {
...
+    implementation("com.google.api.grpc:proto-google-common-protos:2.9.0") // version pulled from protobuf-java
     implementation("com.google.guava:guava:31.1-android") // version pulled from protobuf-java
     compileOnly("javax.annotation:javax.annotation-api:1.3.2")
 
-    permitUsedUndeclared("com.google.api.grpc:proto-google-common-protos:2.9.0")
-
-    protobuf(files("../proto/"))
+    protobuf(files(fileTree("../proto") {
+        include("*.proto")
+        exclude("**/*/*.proto")
```
nand4011, line 35: This is the meat of the change. The top level proto files are ours. There is a google directory containing the protos we want to exclude.

cprice404 (approved):

PR #192, opened 2023-08-15
feat: move permissions to shared, a new token proto

ben-kugler:
- Moved permissions to a new shared folder
- `GenerateAuthToken` contains no new changes, only moving of `Permissions` message
- New `GenerateTempAuthToken` request and response, is the same as `GenerateAuthToken` however it doesn't include a expiration with option of `Never` and response will never include a `refreshToken`


proto/token.proto
```diff
@@ -0,0 +1,37 @@
...
+import "permissions.proto";
+
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_multiple_files = true;
+option java_package = "momento.auth";
+
+package token;
+
+service Token {
+  rpc GenerateTempAuthToken (_GenerateTempAuthTokenRequest) returns (stream _GenerateTempAuthTokenResponse) {}
```
kvcache, line 12:
```suggestion
  rpc GetTemporaryAuthToken (_GetTemporaryAuthTokenRequest) returns (_GetTemporaryAuthTokenResponse) {}
```

* I think you can just use "get" instead of "generate" to mean the same thing.
* We should prefer clear, full words like "temporary" over needless abbreviations like "temp."
* If you think you need a server-side streaming response, let's chat on slack. I'm guessing that was just a mistaken copypaste?

proto/token.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+message _GenerateTempAuthTokenRequest {
```
ben-kugler, line 15: This still needs to be finalized alternative suggestions
- GenerateEphemeralAuthToken
- GenerateProvisionalAuthToken
- GenerateBriefAuthToken
- GeneratePerishableAuthToken
- GenerateDecayingAuthToken
- GenerateTransientAuthToken

proto/token.proto
ben-kugler: - I think you can just use "get" instead of "generate" to mean the same thing.
  -  I like that
- We should prefer clear, full words like "temporary" over needless abbreviations like "temp."
  - Yes, this name also happens to be temporary, this pr is in effort to get more of a discussion/decision on the API name
- If you think you need a server-side streaming response, let's chat on slack. I'm guessing that was just a mistaken copypaste?
  - Yep, def a copypaste miss
  
I'll clean these up once name is settled

proto/token.proto
```diff
@@ -0,0 +1,35 @@
...
+    // generate a token that has an expiry
```
krispraws, line 16: GenerateAuthToken used to take the source token used for the call as an input parameter `auth_token` - I don't remember why. Does this api use the token from the Authorization header?

proto/auth.proto
```diff
@@ -69,98 +71,9 @@ message _GenerateApiTokenRequest {
...
-        }
-        oneof topic {
-          All all_topics = 4;
-          TopicSelector topic_selector = 5;
-        }        
-      }
-    }
-
-    Permissions permissions = 4;
+  permissions.Permissions permissions = 4;
```
krispraws, line 76: Can you try generating the protos for the web sdk and compiling the web sdk with them? we used some hacks there to get it to work with strictjs and I am worried that moving these messages to different packages is going to break it.

proto/permissions.proto
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Aliases for categories of functionality.
+enum TopicRole {
+  TopicPermitNone = 0;
+  // Restricts access to apis that read and write data from topics: No higher level resource description or modification.
+  TopicReadWrite = 1;
+  // Restricts access to apis that read from topics: No higher level resource description or modification.
+  TopicReadOnly = 2;
+  // Only publish allowed
```
krispraws, line 27: Nit: please fix comment to be consistent

proto/auth.proto
krispraws: @cprice404 FYI ^

proto/permissions.proto
```diff
@@ -0,0 +1,96 @@
+syntax = "proto3";
+
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_multiple_files = true;
+option java_package = "momento.shared.permissions";
+
+package permissions;
```
krispraws, line 7: if we can avoid using a package here, that might be sufficient to keep the web sdk working.

proto/token.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+
+import "permissions.proto";
+
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_multiple_files = true;
+option java_package = "momento.auth";
+
+package token;
```
krispraws, line 9: Same for this - do we get any real benefit from using packages for new proto files? grpc-web + strictjs doesn't like them. Also nit: the java package doesn't match.

rust/proto_generator/build.rs
```diff
@@ -15,7 +15,9 @@ fn main() {
         .out_dir(out_dir)
         .compile(
             &[
+                format!("{proto_dir}/permissions.proto"),
```
krispraws, line 18: You need to go and add these files to the generator scripts for each language - we hardcode the names of the files.

proto/permissions.proto
cprice404: I agree that we need to test the web SDK but I don't know that I want to introduce a new inconsistency in our use (or lack thereof) of packages until we know there's a problem with the web SDK. I'd rather be consistent with our current conventions and see if we can solve it for Web SDK.

proto/permissions.proto
```diff
@@ -0,0 +1,96 @@
+syntax = "proto3";
```
cprice404, line 1: is there a diff in this file that we should be paying attention to, or is this just an unchanged migration to a separate file

proto/token.proto
cprice404: i think we aligned in slack that we are not going to switch to `Get`, we will keep `Generate` for consistency with existing API. Jury might still be out on `Temporary` vs other names

proto/token.proto
cprice404: it was because some of the APIs required a session token and others required an API token, and it seemed more clear to make them input parameters so that we weren't overloading the meaning of the header.

Unless we have a good reason to change things up I think I'd vote for consistency again and just say that all Auth APIs accept their required auth tokens as inputs. It's more explicit and less magical.

rust/proto_generator/build.rs
cprice404: good catch Ramya, thank you!

proto/permissions.proto
krispraws: Sure. The hack will almost definitely break though, because even if we remove `package permissions` temporarily, the types in it are still being referred to as `permissions.*` in the other protos.

proto/permissions.proto
ben-kugler: No diff, unchanged migration to a separate file

proto/token.proto
ben-kugler: Ah, this is a good point, the auth clients for our sdk's don't send the token over the auth headers. I was sort of making the assumption that we would be making a new client for this, would want to get @cprice404 thoughts on that.

proto/auth.proto
ben-kugler: Got a successful build without `package permissions`, did have to change imports, here's the `javascript-web` changes
https://github.com/momentohq/client-sdk-javascript/pull/729

image

ben-kugler: Okay, this was a real pain. For `Javascript-web`, I removed the `package` declarations, but in doing so, rust builds will fail since tonic server seems to build protos into the out directory only if they're a package, couldn't find anything in [the docs to help with this](https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.server_attribute). 

Noticing that the `javascript-web` build script actually comments out `package`, I thought I would keep the package declares and  update `permissions.Permissions` to `Permissions` for  `javascript-web` only. It ain't pretty.


Got it working and auth test running in `javascript-web`,  `javascript-nodejs` I was having much worse luck with, `@grpc/grpc-js` is shared between the packages, so `ChannelCredentials, Interceptor` had conflicts between the two different versions.
image


Alternatively to avoid all of this, we could maintain two copies of the permissions object in `auth` and `token`, but not a hug fan of this.

cprice404: the javascript-web build is already a rube goldberg machine. Personally I am okay with pushing more horror and complexity there rather than letting it bleed into other languages/SDKs. If @krispraws is.

krispraws: dumb question - Do we need to put each proto file/each grpc service in its own package? Can tokens.proto, permissions.proto and auth.proto all live in the same `auth` package?

krispraws: I agree with @cprice404 - we can push the complexity into javascript-web builds.

proto/token.proto
krispraws: For better or worse, we have landed on `GenerateAuthAcorn`.

proto/permissions.proto
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  message CacheItemSelector {
+    oneof kind {
+      bytes key = 1;
+      bytes key_prefix = 2;
+    }
+  }
+
+  message CachePermissions {
```
wandaitzuchen, line 67: These message definitions are shared between regular v1 auth and acorn, however we don't actually want to support CacheItemSelector for v1 auth, right? Also, do we want to allow an acorn to be used for topics?
Perhaps it's clearer to have these restrictions in proto instead of mr2?

ben-kugler: For now, yes, the control-plane will reject generate v1 token request that include `CacheItemSelector`, but if we do want to support `CacheItemSelector` in v1 tokens later, i think it will be easier to make validation and sdk changes to allow rather then proto changes. Yes, we want to allow acorn's for us in topics, I think one of the key use cases for acorn's was browser chats.
i do think you're right that it's probably would be more clear, but I think the only groups that will have troubles are the brave souls using our proto's directly.

wandaitzuchen (approved):

PR #193, opened 2023-08-16
refactor: prefix messages with underscore in vector protos

malandis:
Our convention is to prefix message names with an underscore to not pollute a
user's IDE autocomplete. Because our SDKs will publish
response types with names identical or similar to the names in the
protos, we want to distinguish the protobuf-generated names from names
in the SDK. Our way of doing this is prefixing the protobuf message
names with an underscore.


nidhidhamnani (approved):
:shipit:

PR #195, opened 2023-08-17
feat: include cache and topic limits in _ListCachesResponse

tylerburdsall:
Customers don't have a way to know what their current limits are, this exposes them in the proto so that we can vend that infomration

### DIFF
--- a/proto/controlclient.proto
+++ b/proto/controlclient.proto
@@ -1,5 +1,6 @@
 syntax = "proto3";
 
+
 option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
 option java_multiple_files = true;
 option java_package = "grpc.control_client";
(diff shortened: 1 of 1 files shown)


proto/controlclient.proto
```diff
@@ -60,8 +61,30 @@ message _ListCachesRequest {
...
+  uint32 max_throughput_kbps = 2;
+  // The maximum size of a single item in KiB
+  uint32 max_item_size_kb = 3;
+  // The maximum TTL allowed for a single item, in seconds
+  uint64 max_ttl_seconds = 4;
+}
+
+message _TopicLimits {
+  // The amount of messages that can be published per second
+  uint32 max_publish_rate = 1;
```
bruuuuuuuce, line 77: do you know if this is _per_ topic, or across all topics?

tylerburdsall: This is _per cache_, so across all topics

bruuuuuuuce (approved):
:shipit: LGTM

PR #200, opened 2023-08-23
chore: add `_Cache` to the create cache response

bruuuuuuuce:


tylerburdsall (dismissed):

proto/controlclient.proto
```diff
@@ -54,9 +54,6 @@ message _CreateCacheRequest {
   string cache_name = 1;
 }
 
-message _CreateCacheResponse {
-}
-
```
tylerburdsall, line 59: Can we just edit the message in-place here so that `_CreateCacheResponse` comes after `_CreateCacheRequest` and maintains order?

tylerburdsall (commented):
Left a minor nit

proto/controlclient.proto
bruuuuuuuce: yea good call

tylerburdsall (approved):

cprice404 (approved):

rishtigupta (approved):

PR #202, opened 2023-09-01
chore: add missing export for vectorindex namespace in grpc-web

cprice404:


proto/auth.proto
```diff
@@ -6,7 +6,7 @@ option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
 option java_multiple_files = true;
 option java_package = "momento.auth";
 
-package auth;
+//package auth;
```
eaddingtonwhite: Why did we comment out all the package names? JC want to double check meant to leave that in?

malandis (commented):
LGTM but not sure why builds are failing?

pgautier404 (approved):

PR #209, opened 2023-09-21
chore: Add leaderboard definitions

kvcache:
A first step toward a leaderboard offering, I've documented and
translated the internal proposal to a grpc protocol buffers spec.

This isn't connected to any client languages yet; those can be
opted in as we add support and implement the service.

### DIFF
--- a/proto/leaderboard.proto
+++ b/proto/leaderboard.proto
@@ -1,0 +1,192 @@
+syntax = "proto3";
+
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_multiple_files = true;
+option java_package = "grpc.leaderboard";
+option csharp_namespace = "Momento.Protos.Leaderboard";
+
+package leaderboard;
+
+// Like a sorted set, but for leaderboards!
+//
+// Elements in a leaderboard are keyed by an ID, which is an unsigned 64 bit integer.
+// Scores are single-precision floating point numbers.
+//
+// Each ID can have only 1 score.
+//
+// For batchy, multi-element apis, limits are 8192 elements per api call.
+//
+// Scores are IEEE 754 single-precision floating point numbers. This has a few
+// implications you should be aware of, but the one most likely to affect you is that
+// below -16777216 and above 16777216, not all integers are able to be represented.
+service Leaderboard {
+  // Deletes a leaderboard. After this call, you're not incurring storage cost for this leaderboard anymore.
+  rpc DeleteLeaderboard(DeleteLeaderboardRequest) returns (Empty);
+
+  // Insert or update elements in a leaderboard. You can do up to 8192 elements per call.
+  // There is no partial failure: Upsert succeeds or fails.
+  rpc UpsertElements(UpsertElementsRequest) returns (Empty);
+
+  // Remove up to 8192 elements at a time from a leaderboard. Elements are removed by id.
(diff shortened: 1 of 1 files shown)


virratanasangpunth (dismissed):
LGTM

proto/leaderboard.proto
```diff
@@ -0,0 +1,192 @@
...
+message ScoreRange {
+  // IEEE 754 single precision 32 bit floating point number.
+  // Momento does not support NaN or Inf in leaderboards.
+  float score_start = 1;
+  // IEEE 754 single precision 32 bit floating point number.
+  // Momento does not support NaN or Inf in leaderboards.
+  float score_end = 2;
+}
+
+message Empty {}
```
tylerburdsall, line 134: @kvcache have you considered using GRPC's default `Empty` message? That way we don't have to have multiple classpaths for an empty message

tylerburdsall (commented):
Left a comment

proto/leaderboard.proto
```diff
@@ -0,0 +1,192 @@
...
+
+// Query APIs using ScoreRange may match more than the limit of 8192 elements. These apis will
+// include an offset and limit parameter pair, which can be used to page through score ranges
+// matching many elements.
+//
+// ScoreRange models half-open ranges: 0..4 refers to scores 0, 1.1234, 2.5 and 3.999.
+message ScoreRange {
+  // IEEE 754 single precision 32 bit floating point number.
+  // Momento does not support NaN or Inf in leaderboards.
+  float score_start = 1;
```
krispraws, line 128: Nit - naming is inconsistent with RankRange where names also indicate inclusive/exclusive

tylerburdsall (commented):
Also I believe this RPCs Request/Response messages should be prefixed with `_`

krispraws (approved):
Looks good. Left a naming nit comment

proto/leaderboard.proto
kvcache: Yes I did consider that. But this is easier and doesn't have any extra import dependencies. It's also binary-compatible with that message if you ever want to delete this line and replace it with another line that imports this exact structure from someone else's code :-p! This is basically just a more intent-revealing replacement for `XRpcResponse {}` boilerplate that we copypaste and replace X for in many other places.

proto/leaderboard.proto
kvcache: woops thanks

kvcache:
> Also I believe this RPCs Request/Response messages should be prefixed with `_`

Thanks, I forgot this one

tylerburdsall (approved):

krispraws (approved):

proto/leaderboard.proto
```diff
@@ -0,0 +1,192 @@
...
+  // IEEE 754 single precision 32 bit floating point number.
+  // Momento does not support NaN or Inf in leaderboards.
+  float end_exclusive = 2;
+}
+
+message _Empty {}
+
+message _DeleteLeaderboardRequest {
+  string cache_name = 1;
+  string leaderboard = 2;
```
virratanasangpunth, line 138: nit: should `leaderboard` be `leaderboard_name`? I see we're using `set_name` and `list_name` in the cache proto.

kvcache: Chatted separately - this ship has kind of sailed at the moment. We can change it later if desired.

PR #214, opened 2023-09-27
feat: model unbounded score range explicitly

virratanasangpunth:


tylerburdsall (dismissed):

anitarua (commented):
could you also do the same for rank range?

virratanasangpunth:
> could you also do the same for rank range?

@anitarua No, that API is a little different, you can't request "gimme players between rank 0 and infinity". If a client wants that behavior they need to "paginate" manually, e.g.
```
one_page_of_players = [];
all_players = [];
start = 0;
end = 8192;
{
  one_page_of_players = get_by_rank(start, end);
  all_players.push(one_page_of_players);
  start += 8192;
  end += 8192;
} while (one_page_of_players !== [])
```

anitarua (approved):

tylerburdsall (approved):

kvcache (approved):
yep looks right

proto/leaderboard.proto
```diff
@@ -94,18 +94,26 @@ message _RankRange {
...
-  // Momento does not support NaN or Inf in leaderboards.
-  float start_inclusive = 1;
-  // IEEE 754 single precision 32 bit floating point number.
-  // Momento does not support NaN or Inf in leaderboards.
-  float end_exclusive = 2;
+  oneof min {
+    _Unbounded unbounded_min = 1;
+    // IEEE 754 single precision 32 bit floating point number.
+    // Momento does not support NaN or Inf in leaderboards.
+    float min_score = 2;
```
krispraws, line 109: Saw this after it was merged - could we rename `min_score` to `inclusive_min` and `max_score` to `exclusive_max`?

virratanasangpunth: sure.

PR #215, opened 2023-09-27
chore: add inclusive/exclusive suffix

virratanasangpunth:


proto/leaderboard.proto
```diff
@@ -106,13 +106,13 @@ message _ScoreRange {
     _Unbounded unbounded_min = 1;
     // IEEE 754 single precision 32 bit floating point number.
     // Momento does not support NaN or Inf in leaderboards.
-    float min_score = 2;
+    float min_inclusive = 2;
```
krispraws, line 109: `inclusive_min` goes better with `unbounded_min` but 🤷‍♀️

anitarua (approved):

krispraws (approved):

proto/leaderboard.proto
tylerburdsall: I'm with @krispraws on this one tbh

virratanasangpunth: but for RankRange we have start_inclusive/end_exclusive....

tylerburdsall: Oh boy 🙈 Well, in that case this is fine, naming is hard

tylerburdsall (approved):

proto/leaderboard.proto
krispraws: So many ways to model a range 🙈  https://github.com/momentohq/client-protos/blob/375c8f8cfc783ecdb62fea1422428eb96b8978f0/proto/cacheclient.proto#L679

PR #216, opened 2023-09-28
chore: add leaderboard.proto to javascript generate_protos scripts

anitarua:


anitarua:
Going with a two-pass build for js-web instead of renaming the duplicate proto messages

PR #222, opened 2023-10-03
fix: edit python generated pb2 import statements in general

malandis:
In the previous version of `python/run-protoc.sh`, we selectively
edited import statements for `extensions_pb2`. With the prior addition
of `permissionmessages`, the package is broken. We need to edit those
as well.

In order to avoid this in the present and future, we generalize the
edits to all pb2 import statements in our generated code.


nand4011 (approved):

pgautier404 (approved):
Nice, thanks!

PR #224, opened 2023-10-06
chore: use uint32 datatypes instead of uint64 for leaderboards

tylerburdsall:


proto/leaderboard.proto
```diff
@@ -90,8 +90,8 @@ message _RankedElement {
 // * 4..3
 // * 0..8193
 message _RankRange {
-  uint64 start_inclusive = 1 [jstype = JS_STRING];
-  uint64 end_exclusive = 2 [jstype = JS_STRING];
```
tylerburdsall, line 94: cc @anitarua I had to remove these annotations, it looks like that `jstype` is only for 64-bit type numbers, I believe JS should handle the uint32 type natively now

virratanasangpunth (approved):

anitarua (approved):

PR #226, opened 2023-10-17
feat: add more vector index item metadata value types

malandis:
Expands vector index item metadata value types to allow not just
string but:
- integer (int64)
- double (64-bit precision float)
- boolean
- list of strings

These values cover commonly used metadata types.

### DIFF
--- a/proto/vectorindex.proto
+++ b/proto/vectorindex.proto
@@ -37,12 +37,17 @@ message _Vector {
 
 message _Metadata {
     string field = 1;
+
+    message ListOfStrings {
+        repeated string values = 1;
+    }
+
     oneof value {
         string string_value = 2;
-	int64 integer_value = 3;
-	double double_value = 4;
-	bool boolean_value = 5;
-	repeated string list_of_strings_value = 6;
+        int64 integer_value = 3;
+        double double_value = 4;
+        bool boolean_value = 5;
+        ListOfStrings list_of_strings_value = 6;
     }
 }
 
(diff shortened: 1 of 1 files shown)


proto/vectorindex.proto
```diff
@@ -37,9 +37,17 @@ message _Vector {
...
+        repeated string values = 1;
+    }
+
     oneof value {
-        // Eventually can support ints, dates, etc
         string string_value = 2;
+        int64 integer_value = 3;
+        double double_value = 4;
+        bool boolean_value = 5;
+        ListOfStrings list_of_strings_value = 6;
```
honeyAcorn, line 50: I had this instead `_ListOfStrings list_of_strings_value = 6;`

Following what other message naming

proto/vectorindex.proto
```diff
@@ -37,9 +37,17 @@ message _Vector {
 
 message _Metadata {
     string field = 1;
+
+    message ListOfStrings {
```
honeyAcorn, line 41: I had `_ListOfString`

proto/vectorindex.proto
malandis: I can change that, though it seems to be inconsistent in this file, eg
image

malandis: Fixed in https://github.com/momentohq/client-protos/pull/226/commits/f88b92d6cc4b5bfc8a1f257fa71d817035b57b5c

PR #231, opened 2023-10-30
feat: add search radius to filter results

honeyAcorn:
We are adding search threshold (radius) to filter the search results.

### DIFF
--- a/proto/vectorindex.proto
+++ b/proto/vectorindex.proto
@@ -64,11 +64,17 @@ message _MetadataRequest {
     }
 }
 
+message _NoScoreThreshold {}
+
 message _SearchRequest {
     string index_name = 1;
     uint32 top_k = 2;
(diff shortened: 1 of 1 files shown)


malandis (commented):
Since `_NoScoreThreshold` is common to both search request messages, we can factor that out. Otherwise looks good!

honeyAcorn:
> Since `_NoScoreThreshold` is common to both search request messages, we can factor that out. Otherwise looks good!

Is it okay to keep it as is cause I already merged the backend.  https://github.com/momentohq/mr2rs/blob/7956d7f171f62e9f2f6a5a1762a75751a42db6ea/proto/vectorindex.proto#L68

I'll create another PR to factor that out in our backend. Would that be okay?

malandis (approved):

PR #232, opened 2023-10-31
feat: GetItemBatch API proto

honeyAcorn:
This will allow us to get item by specifying all id. 
We do GetItemBatch API first. 
Then will have another PR to do the GetItemAndFetchVectorsBatch later.

### DIFF
--- a/proto/vectorindex.proto
+++ b/proto/vectorindex.proto
@@ -98,3 +98,25 @@ message _SearchResponse {
 message _SearchAndFetchVectorsResponse {
     repeated _SearchAndFetchVectorsHit hits = 1;
 }
+
+message _GetItemBatchRequest {
+  string index_name = 1;
+  repeated string ids = 2;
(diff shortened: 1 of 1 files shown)


honeyAcorn:
Do we want to let _Miss also return the original requested id?
Otherwise user will just have to order it one by one, not sure about the use case.
In other words, if we have id1, and id2 in index, but not id3, when we request it like this:
image
for id3 it will just show `Miss`, is that okay?

malandis:
> for id3 it will just show Miss, is that okay?

Yes that's OK.

As long as the service orders the hit or miss results identically to the requested ids, we don't need to repeat the ids in the `Miss`. The client libraries can take of packaging up the results in a nice way for the application developer.

cprice404 (approved):
lgtm if michael is happy

malandis (approved):

PR #238, opened 2023-11-08
feat: return secret_string along in webhook body for listResponse

ben-kugler:
We need to add `secret_string` as part of the response for `_ListWebhooksResponse`

Options
1. First commit, add optional `secret_string` to `_Webhook`, this message is used in `_PutWebhookRequest` and `_ListWebhooksResponse`, so might be a little odd/confusing that a secret string can be supplied on put webhook request.
2. Second commit, wrapper for webhook and secret_string, `_WebhookWithSecret` for `_ListWebhooksResponse`

```
message _Webhook {
  _WebhookId webhook_id = 1;
  string topic_name = 2;
  _WebhookDestination destination = 3;
}

message _WebhookWithSecret {
  _Webhook webhook = 1;
    string secret = 2;
}

message _ListWebhooksResponse {
  repeated _WebhookWithSecret webhook = 1;
}

### DIFF
--- a/proto/webhook.proto
+++ b/proto/webhook.proto
@@ -50,7 +50,7 @@ message _ListWebhookRequest {
 }
 
 message _ListWebhooksResponse {
-  repeated _WebhookItem webhook_secret = 1;
+  repeated _WebhookItem webhook_item = 1;
 }
 
 message _WebhookDestination {
--- a/proto/webhook.proto
+++ b/proto/webhook.proto
@@ -23,6 +23,9 @@ message _Webhook {
   _WebhookId webhook_id = 1;
   string topic_name = 2;
   _WebhookDestination destination = 3;
+  oneof secret_string {
+    string secret = 4;
+  }
 }
 
 message _PutWebhookRequest {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by bruuuuuuuce — APPROVED** [review]
:shipit: looks clean to me 🧹

File: proto/webhook.proto [hunk 1]
```diff
@@ -25,6 +25,11 @@ message _Webhook {
   _WebhookDestination destination = 3;
 }
 
+message _WebhookSecret {
```
virratanasangpunth, line 28: `_WebhookItem` or something? That way the name is not tied to what fields are declared in the struct. Not a strong opinion. Either way the names that will show in the SDKs will matter more.

virratanasangpunth (dismissed):

proto/webhook.proto
```diff
@@ -45,7 +50,7 @@ message _ListWebhookRequest {
 }
 
 message _ListWebhooksResponse {
-  repeated _Webhook webhook = 1;
+  repeated _WebhookItem webhook_secret = 1;
```
virratanasangpunth, line 53:
```suggestion
  repeated _WebhookItem webhook_item = 1;
```

virratanasangpunth (approved):

PR #239, opened 2023-11-08
feat: add detail to MVI list indexes RPC

malandis:
Previously the list indexes RPC only returned a list of index
names. This PR modifies this to return a list of index detail
containing index name, num dimensions, and similarity metric.

We also factor out the similarity metric message from the
`CreateIndexRequest` since it is common to create index, list indexes,
and (in the future) describe index.

Because the service is in early beta we have carefully weighed
changing this and have agreed the earlier the better.

### DIFF
--- a/proto/controlclient.proto
+++ b/proto/controlclient.proto
@@ -21,19 +21,27 @@ service ScsControl {
   rpc ListIndexes(_ListIndexesRequest) returns (_ListIndexesResponse) {}
 }
 
+message _SimilarityMetric {
+  message _EuclideanSimilarity {
+  }
+
+  message _InnerProduct {
+  }
+
+  message _CosineSimilarity {
+  }
+
+  oneof similarity_metric {
+    _EuclideanSimilarity euclidean_similarity = 1;
+    _InnerProduct inner_product = 2;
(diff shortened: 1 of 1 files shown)


bruuuuuuuce (approved):
:shipit:

honeyAcorn (approved):
LGTM! Please hold off the Merge until backend ready! Thanks!

proto/controlclient.proto
```diff
@@ -21,21 +21,29 @@ service ScsControl {
   rpc ListIndexes(_ListIndexesRequest) returns (_ListIndexesResponse) {}
 }
 
-message _CreateIndexRequest {
-  message _EuclideanSimilarity{}
+message _SimilarityMetric {
+  message _EuclideanSimilarity {
+  }
```
honeyAcorn, line 26: Any reason to separate closing brace } to next line?

cprice404 (approved):
❤️ ty

proto/controlclient.proto
malandis: In the rest of the file, empty messages are closed on the next line. I just glanced at the rest of the file and followed that style.

PR #240, opened 2023-11-08
chore: mvi - change distance field to score to align with similarity

malandis:
The SearchHit and SearchAndFetchVectorHit structures currently include
a distance field, which is misleading and inconsistent with our naming
conventions. To clarify:

(1) Our Search RPCs utilize a score_threshold to filter results,
implying the evaluation of scores rather than distances.

(2) The term distance suggests that a lower value is preferable,
whereas our system employs similarity metrics where a higher value
signifies better similarity. For example, cosine similarity is a
common metric we use, where a higher score indicates greater
similarity.

To avoid confusion and maintain consistency, we are renaming the
distance field to score. This change is a simple renaming and does not
alter the representation on the wire, ensuring that existing
integrations remain functional.

### DIFF
--- a/proto/vectorindex.proto
+++ b/proto/vectorindex.proto
@@ -91,13 +91,13 @@ message _SearchAndFetchVectorsRequest {
 
 message _SearchHit {
     string id = 1;
-    float distance = 2;
+    float score = 2;
     repeated _Metadata metadata = 3;
 }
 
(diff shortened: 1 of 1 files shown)


honeyAcorn (approved):
make sense! Approve now but let's not merge it until I had backend updated!

PR #242, opened 2023-11-09
feat: add new get secret endpoint for webhooks

ben-kugler:


proto/webhook.proto
```diff
@@ -48,6 +54,15 @@ message _ListWebhooksResponse {
   repeated _Webhook webhook = 1;
 }
 
+message _DescribeWebhookRequest {
+  string cache_name = 1;
+  string webhook_name = 2;
+}
+
+message _DescribeWebhookResponse {
+  repeated _WebhookItem webhook_item = 1;
```
virratanasangpunth, line 63: I don't think this is `repeated`.

IMO let's just return these 3 fields: `cache_name`,`webhook_name`,`secret_string` here.

ben-kugler: lol oops, copy-paste.

Should this not just be called `DescribeWebhookSecret` if the only new information we're returning is a secret?

virratanasangpunth: Oh yea. That too. This should be called `DescribeWebhookSecret`

ben-kugler: Another option (that I'm more partial for) is not adding the secret_string outside of `Put` for now, I think there some stuff here we might want to think about a bit more, like Chris said, inclusion of teams/orgs. 

For example, what if we only gave access to the webhook secrets through a pi on the `tokens` endpoint? 

I think also, I don't even know why we are doing this? To display/get secrets on the console? Does that need to be in MVP?

proto/webhook.proto
```diff
@@ -25,6 +26,11 @@ message _Webhook {
   _WebhookDestination destination = 3;
 }
 
+message _WebhookItem {
+  _Webhook webhook = 1;
+  string secret = 2;
+}
+
```
virratanasangpunth, line 33: this is not used anywhere?

virratanasangpunth (approved):

bruuuuuuuce (approved):
:shipit:

PR #243, opened 2023-11-10
feat: add ListIndex detail

honeyAcorn:
Currently, when we called `listIndex`, it only returns the name of indexes. We have not provided a way for customer to know about other index details. Therefore, we want to include `number of dimensions` and `similarity metric type` when calling listIndex


honeyAcorn:
Forgot @malandis already have it. Closing this one

PR #245, opened 2023-11-20
feat: add Kotlin Android and JVM protos

nand4011:
Copy configuration from the grpc kotlin stub lite example to generate kotlin code for android and jvm.

Remove old kotlin messages generation and publication.

I needed to include google/protobuf/descriptor.proto because there is no lite version of google common protos:
https://github.com/protocolbuffers/protobuf/issues/7331
https://github.com/protocolbuffers/protobuf/issues/1889

> There's no proto-google-common-protos-lite because there is no API/ABI stability guarantees for lite codegen. You need to generate all lite protos with the same version of the generator, so pre-built libraries doesn't make much sense. You can't ship a library that has pre-built lite protos, unless you can control the version of protobuf the app uses (which could be the case within a single company, but not for anything that goes on Maven Central). If you use protobuf only internally to your library, you can shade protobuf, but obviously that bloats its size so isn't really much of a help.

That comment was made last week, so this is an ongoing issue. I don't see a good way of guaranteeing that libraries that use java-lite/kotlin-lite won't clash with each other other than trying to shade the generated code.


kotlin/android/build.gradle.kts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+dependencies {
+    // Descriptor.proto is included because there is no lite version of common protos:
+    // https://github.com/protocolbuffers/protobuf/issues/1889
+    protobuf(files("./google-protos"))
+    protobuf(files(fileTree("../../proto") {
+        include("*.proto")
+        // webhook is temporarily excluded because the kotlin code generation creates non-compiling code
+        // in the case of a proto with a package that shares a name with a message field.
```
pratik151192, line 14: LGTM but once we have a ticket we should prolly link here

kotlin/android/build.gradle.kts
```diff
@@ -0,0 +1,104 @@
+plugins {
+    kotlin("jvm")
+    id("com.google.protobuf")
+    id("maven-publish")
+}
+
+dependencies {
+    // Descriptor.proto is included because there is no lite version of common protos:
+    // https://github.com/protocolbuffers/protobuf/issues/1889
+    protobuf(files("./google-protos"))
```
pratik151192, line 10: Is this the one that includes `descriptor.proto`? Not sure if the google-protos lib was auto generated due to this.. am thinking if we should only pull `descriptor.proto` if at all possible?

nand4011: We need all of the protos in there. The three in `api` are redundant, since they are in the main protos directory, but I couldn't get multiple file includes to work.

cprice404:
@nand4011 are the issues in the original PR description resolved?

kotlin/android/build.gradle.kts
```diff
@@ -0,0 +1,124 @@
...
+
+val javadocJar = tasks.register("javadocJar", Jar::class.java) {
+    archiveClassifier.set("javadoc")
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("maven") {
+            from(components["java"])
+            artifactId = rootProject.name + "-android"
```
cprice404, line 86: what do the final coordinates end up looking like?

nand4011: `software.momento.kotlin:client-protos-jvm` and `software.momento.kotlin:client-protos-android`. I matched the existing pattern for the java protos: `software.momento.java:client-protos`.

nand4011:
@cprice404 Both issues are still open. There has been no comment on the webhook bug I reported with the protobuf project. I was hoping that would be fast enough that it would be fixed before releasing this, but no dice.

If necessary, I can leave the old kotlin protos in place. They are under a different module and they can still be published to the internal repo without disturbing this work.

cprice404:
> @cprice404 Both issues are still open. There has been no comment on the webhook bug I reported with the protobuf project. I was hoping that would be fast enough that it would be fixed before releasing this, but no dice.
> 
> If necessary, I can leave the old kotlin protos in place. They are under a different module and they can still be published to the internal repo without disturbing this work.

I don't think you need to maintain the old version, I just need to understand the implications here before we merge.

It sounds like:

(a) we had to vendor some google protos, and maybe if other projects are trying to use our SDK and also use other grpc libraries, there may be a conflict between the 2 versions of said protos? and
(b) I'm not sure understand the current state of this PR w/rt the "webhook" collision. Did you just skip the webhook APIs in the kotlin proto generation in this PR?

kotlin/android/build.gradle.kts
nand4011: It looks like we don't need the class that is generating with a syntax error, so there is no issue.

kotlin/android/build.gradle.kts
nand4011: I updated this directory to only contain `descriptor.proto`. We don't need to include the others now that we don't need to exclude webhook.

nand4011:
I discovered that we don't need the kotlin version of the webhook class, so there is no issue there anymore. I updated the pr.
The `descriptor.proto` and other google protos we included are still a potential issue. A user that is making an android project that includes multiple libraries that make grpc calls could run into problems. I need to experiment to make sure it is an issue, but it would only affect a narrow slice of users.

cprice404 (approved):

PR #255, opened 2024-01-16
feat: allow both f32 and f64 scores for leaderboard

virratanasangpunth:


cprice404 (dismissed):

anitarua (dismissed):

anp13 (commented):
May be we should add the deprecated flag to the f32 fields like we did in the interserver protos: https://github.com/momentohq/kotlin-kitchensink/pull/535/files

anp13 (approved):

PR #260, opened 2024-02-01
feat: add protos for SetSample

tylerburdsall:


kvcache (approved):

proto/cacheclient.proto
```diff
@@ -338,6 +339,23 @@ message _SetFetchResponse {
...
+}
+
+message _SetFetchRandomResponse {
+  message _Found {
+    repeated bytes elements = 1;
+  }
+
+  message _Missing {}
+
+  oneof set {
```
honeyAcorn, line 354: sorry I just notice that we call it `set` here and I already approved it. But will it be more clear if we use result? Since the set elements is technically inside _Found and result is either found or missing? I'm still approving this because I should have brought up my concern earier.

honeyAcorn (approved):

kvcache (dismissed):
thanks for updating the limit name

honeyAcorn (approved):

honeyAcorn (approved):

nidhidhamnani (approved):

kvcache (approved):

PR #271, opened 2024-03-04
chore: move common messages to common.proto

pgautier404:
This commit refactors the `Unbounded` and `Empty` messages out to `common.proto`. This was motivated by build failures in the golang SDK, where the multiply declared messages were flagged as having been redefined because all messages live in the same package namespace.


tylerburdsall (approved):
Awesome, thanks for this!

honeyAcorn (approved):

PR #273, opened 2024-03-07
chore: add common.proto to javascript-web

pgautier404:
This commit updates the list of files to compile with `protoc` to include `common.proto`, which contains messages shared among other protos. The preprocessing of protos specific to js-web is also updated to remove the package qualifiers from messages in `common.proto` in the same fashion that `permission_messages.Permissions` is being handled. Finally, a second `protoc` pass that was necessary to accommodate messages duplicated across protos has been removed, as the common messages are now shared instead of being re-declared in multiple places.


anitarua (approved):
nice!

bruuuuuuuce (approved):
:shipit:

krispraws (approved):

PR #275, opened 2024-04-02
chore: bump grpc js

pratik151192:
Contains a fix for a related issue we are seeing during service deployments https://github.com/grpc/grpc-node/issues/2625. Will test locally but having issues with local cell so don't want to block on the upgrade.


bruuuuuuuce (approved):
:shipit:

PR #277, opened 2024-04-17
feat: add store proto

honeyAcorn:
New Store proto for Persistent Store

### DIFF
--- a/proto/store.proto
+++ b/proto/store.proto
@@ -13,8 +13,8 @@ service Store {
   rpc Delete(_DeleteRequest) returns (_DeleteResponse) {}
 }
 
-message _ValueType {
-  oneof type {
+message _Value {
+  oneof value {
     bytes bytes_value = 1;
     string string_value = 2;
     int64 integer_value = 3;
@@ -28,20 +28,13 @@ message _GetRequest {
 }
 
 message _GetResponse {
-  oneof result {
-    Miss miss = 1;
-    Hit hit = 2;
-  }
-  message Miss { }
-  message Hit {
-    _ValueType value = 3;
-  }
+    _Value value = 1;
 }
 
 message _SetRequest {
   string store_name = 1;
   bytes key = 2;
-  _ValueType value = 3;
+  _Value value = 3;
 }
 
 message _SetResponse { }
--- a/proto/store.proto
+++ b/proto/store.proto
@@ -13,9 +13,16 @@ service Store {
   rpc Delete(_DeleteRequest) returns (_DeleteResponse) {}
 }
 
+message _ValueType {
+  bytes bytes_value = 3;
+  string string_value = 4;
+  int64 integer_value = 5;
+  float float_value = 6;
+}
+
 message _GetRequest {
-  bytes key = 1;
-  string store_id = 2;
+  string store_name = 1;
+  bytes key = 2;
 }
 
 message _GetResponse {
@@ -25,21 +32,21 @@ message _GetResponse {
   }
   message Miss { }
   message Hit {
-    bytes value = 1;
+    _ValueType value = 3;
   }
 }
 
 message _SetRequest {
-  bytes key = 1;
-  bytes value = 2;
-  string store_id = 3;
+  string store_name = 1;
+  bytes key = 2;
+  _ValueType value = 3;
 }
 
 message _SetResponse { }
 
 message _DeleteRequest {
-  bytes key = 1;
-  string store_id = 2;
+  string store_name = 1;
+  bytes key = 2;
 }
 
 message _DeleteResponse { }
--- a/proto/store.proto
+++ b/proto/store.proto
@@ -14,10 +14,12 @@ service Store {
 }
 
 message _ValueType {
-  bytes bytes_value = 1;
-  string string_value = 2;
-  int64 integer_value = 3;
-  double double_value = 4;
+  oneof type {
+    bytes bytes_value = 1;
+    string string_value = 2;
+    int64 integer_value = 3;
+    double double_value = 4;
+  }
 }
 
 message _GetRequest {
--- a/proto/store.proto
+++ b/proto/store.proto
@@ -1,0 +1,45 @@
+syntax = "proto3";
+
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_multiple_files = true;
+option java_package = "grpc.store";
+option csharp_namespace = "Momento.Protos.Store";
+
+package store;
+
+service Store {
+  rpc Get(_GetRequest) returns (_GetResponse) {}
+  rpc Set(_SetRequest) returns (_SetResponse) {}
+  rpc Delete(_DeleteRequest) returns (_DeleteResponse) {}
+}
+
+message _GetRequest {
+  bytes key = 1;
+  string store_id = 2;
+}
+
+message _GetResponse {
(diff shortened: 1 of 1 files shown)


proto/store.proto
```diff
@@ -0,0 +1,45 @@
...
+
+service Store {
+  rpc Get(_GetRequest) returns (_GetResponse) {}
+  rpc Set(_SetRequest) returns (_SetResponse) {}
+  rpc Delete(_DeleteRequest) returns (_DeleteResponse) {}
+}
+
+message _GetRequest {
+  bytes key = 1;
+  string store_id = 2;
```
anp13, line 18: we don't need a `store_id` for client facing protos. We will need the `store_name` I think but I am not sure because in the cache get/set/delete apis we don't pass the cache_name.

@cprice404 could you comment on this?

proto/store.proto
```diff
@@ -0,0 +1,45 @@
...
+  }
+  message Miss { }
+  message Hit {
+    bytes value = 1;
+  }
+}
+
+message _SetRequest {
+  bytes key = 1;
+  bytes value = 2;
```
anp13, line 34: Following the same naming convention as our cache apis this should be:
```
bytes store_key = 1;
bytes store_value = 2;
```
But I am not sure if we want to name the value as `store_value` or `store_body` as we call it `cache_body` in the cache apis. @cprice404 thoughts?

proto/store.proto
PDXKimani: That rides in a header on cache APIs. That does not mean that is what we should do here, but that's how it works there. I'd personally prefer it to be in the proto, but if Chris has other thoughts I'm open to them.

honeyAcorn:
So we have some options:
`store_id` vs `store_name` or even remove it?
`key` vs `store_key`
`value` vs `store_value` or `store_body`

If we want to keep the naming convention as cacheclient, 
Then it should be `store_key` and `store_body`, and no `store_name`? But from Dylan's reply I think having `store_name` won't hurt.

proto/store.proto
cprice404: tbh I'd prefer it to be in the proto too. But I feel like someone must have had some motivation for putting the cache name in the header originally. @kvcache ?

If @kvcache doesn't remember a compelling reason for why it was in the header for cache then I'm #TeamProto

proto/store.proto
cprice404: idk where `body` came from, `value` makes sense to me.

honeyAcorn: I think `body` came from `cache_body` in cacheclient.proto

cprice404: sorry, what i'm saying is that I don't know how we ultimately chose `body` in the cache proto. `value` makes more sense to me. And overall I do not feel compelled to make this proto match the cache proto, I'd rather just take everything we've learned and make this one the best it can be.

proto/store.proto
kvcache: There were compelling reasons but they no longer exist. Putting it in the message is better in this case, and most cases.

proto/store.proto
```diff
@@ -0,0 +1,45 @@
...
+}
+
+message _GetResponse {
+  oneof result {
+    Miss miss = 1;
+    Hit hit = 2;
+  }
+  message Miss { }
+  message Hit {
+    bytes value = 1;
```
cprice404, line 28: do we want to explicitly model other data types here while we have the chance? There has been a lot of consternation about how in the cache proto we didn't model numeric values in the proto and that forced us to do weird things with them, like requiring them to be strings and then calling increment on strings. I think that if we had to do over with again we probably would have done a `oneof` for the values, and supported string/bytes/int/float as discrete types. cc: @kvcache

honeyAcorn: like this? https://github.com/momentohq/client-protos/pull/277/commits/fff5d408ea9d3d7d16bc46e3359ec5087051d19a

proto/store.proto
honeyAcorn: okay https://github.com/momentohq/client-protos/pull/277/commits/fff5d408ea9d3d7d16bc46e3359ec5087051d19a

proto/store.proto
```diff
@@ -0,0 +1,52 @@
...
+  rpc Set(_SetRequest) returns (_SetResponse) {}
+  rpc Delete(_DeleteRequest) returns (_DeleteResponse) {}
+}
+
+message _ValueType {
+  bytes bytes_value = 1;
+  string string_value = 2;
+  int64 integer_value = 3;
+  double double_value = 4;
+}
```
kvcache, line 21:
```suggestion
message _ValueType {
  oneof {
    bytes bytes_value = 1;
    string string_value = 2;
    int64 integer_value = 3;
    double double_value = 4;
  }
}
```

proto/store.proto
```diff
@@ -0,0 +1,54 @@
...
+
+message _ValueType {
+  oneof type {
+    bytes bytes_value = 1;
+    string string_value = 2;
+    int64 integer_value = 3;
+    double double_value = 4;
+  }
+}
```
anp13, line 23: I would call this message just `Value`
```
message _Value {
  oneof value {
    bytes bytes_value = 1;
    string string_value = 2;
    int64 integer_value = 3;
    double double_value = 4;
  }
}
```

Or may be we can directly use oneof in the request and response like this:
```
message _SetRequest {
  bytes key = 1;
  oneof value {
    bytes bytes_value = 2;
    string string_value = 3;
    int64 integer_value = 4;
    double double_value = 5;
  }
}

message _GetResponse {
  oneof value {
    bytes bytes_value = 1;
    string string_value = 2;
    int64 integer_value = 3;
    double double_value = 4;
  }
}
```

honeyAcorn: Initially I was separating like this, but then I thought, maybe it makes more sense to define them outside since we don't want these two to have chance to diverge. What do you think?

honeyAcorn:
@cprice404 , @kvcache , @anp13 and I had a meeting to talk about this. Here's the summary:

1. We keep the store_name in proto. 
2. We will change the `bytes key` to `string key` for more clarity
3. We will not need hit/miss for `GET` cause this is not like cache, store don't have ttl concept. We will throw `NOT FOUND` grpc error if the key doesn't exist.
4. We will first try to change to `bytes value` to oneof `value` where value can be of `int/string/double/bytes` type. Once it make sense on the sdk and on the server side, we will stick to it and support more type if needed. This would allow us to do operations to numeric types directly without having to convert from string. 

Here's the changes: https://github.com/momentohq/client-protos/pull/277/commits/4715f4fdb228c3db3dee837c12c1c0bcf7d3e9c3

kvcache:
@honeyAcorn 
1 - we flip flopped on this. Store name belongs in the header.
2 - would you like to do that now?
3 - yep.
4 - yep.

honeyAcorn:
> @honeyAcorn 1 - we flip flopped on this. Store name belongs in the header. 2 - would you like to do that now? 3 - yep. 4 - yep.

@kvcache Thanks for correcting it for me! Now it's all good to be reviewed! It looks nice and clean now :) https://github.com/momentohq/client-protos/pull/277/commits/7b8fdf22b3626da20934e954798d860491d495c0

kvcache (approved):
This looks good to me.

cprice404 (approved):
shot

PR #281, opened 2024-05-29
chore: add prefix to store service messages

pgautier404:
This commit prefixes the names of the messages in `store.proto` with the string "Store" to resolve name collisions in the Go SDK. In that SDK all of the messages end up living in the same namespace, so all of the request and response messages (e.g. `_GetRequest`) were colliding with the names in `cacheclient.proto`.

@honeyAcorn @anp13 I wanted to make sure you were both aware of and in agreement with this change. Please let me know if you see any issues or problems with renaming these.

### DIFF
--- a/proto/global_admin.proto
+++ b/proto/global_admin.proto
@@ -1,5 +1,10 @@
 syntax = "proto3";
 
+option go_package = "github.com/momentohq/client-sdk-go;client_sdk_go";
+option java_multiple_files = true;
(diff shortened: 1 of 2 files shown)


kvcache (approved):
This seems fine to me.

anp13 (approved):

bruuuuuuuce (approved):
:shipit:

PR #284, opened 2024-06-06
chore: cache_name will be provided in header rather than proto

honeyAcorn:


proto/leaderboard.proto
```diff
@@ -130,64 +130,64 @@ enum _Order {
 }
 
 message _DeleteLeaderboardRequest {
-  string cache_name = 1;
   string leaderboard = 2;
+  reserved 1;
```
cprice404, line 134: idk if we have a best practice around this, but I wonder if, for historical purposes, it would be useful to leave comments when we make these reservations, so we know what the fields were originally allocated for?

PR #290, opened 2024-06-24
feat: add account session token API

PDXKimani:
Adds a new global API, used for generating account-scoped session tokens.

### DIFF
--- a/proto/global_admin.proto
+++ b/proto/global_admin.proto
@@ -9,6 +9,7 @@ package global_admin;
 
 service GlobalAdmin {
   rpc GetAccounts(_GetAccountsRequest) returns(_GetAccountsResponse) {}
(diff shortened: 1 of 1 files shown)


proto/global_admin.proto
```diff
@@ -23,3 +24,11 @@ message _Account {
   string id = 1;
   string account_name = 2;
 }
+
+message _GetAccountSessionTokenRequest {
+  string account_id = 1;
```
rlinehan, line 29: The customer id/subject will be in the auth context, right, because it'll be from the auth0 token? Are we planning on using the user id at all? Or not yet?

PDXKimani: The Auth0 token establishes the user - the flow is that the console will first call the "GetAccounts" API to get a list of accounts, then pass back the user's selection via this API to generate an auth token scoped to that account.

rlinehan (approved):

PR #292, opened 2024-07-31
feat: add trailers proto

ben-kugler:


proto/trailers.proto
```diff
@@ -0,0 +1,34 @@
...
+    // to complete it either way. A resource was exhausted which prevented the request
+    // from completing.
+    server_is_busy = 1;
+    // Indicates that the stored type for the key supplied isn't compatiable with the 
+    // requested operation
+    invalid_type = 2;
+    // Indicates the item doesn't exist for the given key
+    item_not_found = 3;
+    // Indicates the store doesn't exist
+    store_not_found = 4;
```
malandis, line 33: Will these trailers be used across many services, or are they be specific to the storage client? I ask since the enum variant here is store-specific but the file is called (non-specific) `trailers`.

ben-kugler: These trailers will be used in many services, we are just starting this pattern with the addition of the new store service.

malandis (dismissed):

proto/trailers.proto
```diff
@@ -0,0 +1,34 @@
...
+package trailers;
+
+// Found on the `retry_disposition` trailer for error responses, when the value is known by the service
+enum RetryDisposition {
+     // This rpc is safe to retry, even if it is non-idempotent. It was not executed by the server.
+    retryable = 0;
+    // This rpc may be safe to retry, but it may have been applied.
+    // Non-idempotent commands should not be retried, unless configured explicitly.
+    // Idempotent commands should be considered eligible for retry.
+    possibly_retryable = 1;
```
cprice404, line 16: is there any case where we would want to have a `not_retryable` here, or where the server would know that at the time it created the response?

proto/store.proto
```diff
@@ -26,6 +26,11 @@ message _StoreGetRequest {
   string key = 1;
 }
 
+// This response is for when a get request concludes successfully.
+// These are some of the Errors and their corresponding GRPC status codes.
```
cprice404, line 30: if you are actually going to use the proto for the errors, you should mention it in these comments. As-is this might be interpreted to indicate that the metadata values would be strings (which they currently are).

proto/trailers.proto
cprice404: So we would be adding service-specific (e.g. store_not_found) errors to a shared proto that could be used by other services besides the storage service?

Not that I have an overwhelming objection to that but just want to make sure we are all clear/aligned.

proto/trailers.proto
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// Found on the `err` trailer for error responses, when the error is sent by the server and it has a relevant value to set
+enum Err {
```
cprice404, line 20: I think I like this idea but it will be a breaking change until the SDKs are updated, right?

ben-kugler: Given that we are using these as string's and not the binary protobuf format, yes, changing a metadata name server side would be breaking for SDKs. If that's what you mean, I may be lacking some sdk insight.

proto/trailers.proto
cprice404: okay. just keep in mind that once the sdks are updated to use these, then any changes you make on the server will require additional sdk updates.

proto/trailers.proto
cprice404: > not the binary protobuf format

Wait, I'm confused. If we're not going to use the binary protobuf format then why are we defining these as binary protos?

Since you are introducing these protos in this PR I was assuming that meant that you were going to start using them as the actual values for the `err` metadata. Since the SDKs are currently expecting strings for those, then, as I assumed this PR implies we will be changing from strings to binary protos, then the SDKs would need to be updated and there would be a period of time in between where they were broken.

If you are still planning to send strings back for the `err` metadata then I'm a little confused as to what this proto is for.

cprice404: @ben-kugler if it would be easier to get aligned on this with a short call with a few folks, happy to do that!

ben-kugler: Yeah! That would be nice, I'll get that going. Also, some extra context, the initial idea was to just include trailer warnings as comments, when doing this I then took it a step further to add them as a proto. I do think we should take it further and use the protobuf format if we are going to include them as proto's as you've pointed out. Not doing so would be confusing. If we decide that we aren't going to go in that direction, maybe removing the `trailers` proto would be best.

cprice404: cool. I think it'd be great to have @malandis, @kvcache and @virratanasangpunth in that meeting if possible as they've all been involved in earlier rounds of convo about this.

proto/store.proto
```diff
@@ -7,6 +7,25 @@ option csharp_namespace = "Momento.Protos.Store";
 
 package store;
 
+// Possible headers returned in grpc error responses
+// {
+//   Key: "err"
+//   Values: [
+//    "momento_general_err",
```
malandis, line 14: Could you add back in the clarifying comments for each of the alternatives we had previously? (as in https://github.com/momentohq/client-protos/pull/292/commits/6a6c3cf5e73710ca1f86b9f1668d7a3f0b582ff2).

cprice404:
this wfm if it works for @malandis , @kvcache , and @virratanasangpunth

cprice404 (approved):

malandis (approved):

PR #295, opened 2024-09-23
feat: add invite member apis

rishtigupta:


PDXKimani (approved):

proto/global_admin.proto
```diff
@@ -15,6 +15,11 @@ service GlobalAdmin {
   rpc ListMembers(_ListMembersRequest) returns(_ListMembersResponse) {}
   rpc GetEndpointsForAccount(_GetEndpointsForAccountRequest) returns (_GetEndpointsForAccountResponse) {}
   rpc SetAccountName(_SetAccountNameRequest) returns (_SetAccountNameResponse) {}
+  rpc InviteMember(_InviteMemberRequest) returns (_InviteMemberResponse) {}
+  rpc AcceptInvitation(_AcceptInvitationRequest) returns (_AcceptInvitationResponse) {}
+  rpc RejectInvitation(_RejectInvitationRequest) returns (_RejectInvitationResponse) {}
```
cprice404, line 20: @PDXKimani FWIW I think in the console UI we should use the word "Decline" for this. I don't have any concerns about the proto not matching the console UI if you've already shipped but just calling it out.

cprice404 (approved):

proto/global_admin.proto
PDXKimani: Sure, makes sense.

PR #298, opened 2024-09-30
chore: add topic sequence_page

kvcache:
Add a sequence_page field to topics. This will be in the discontinuities, the subscribe message, and the items next to the sequence number.

### DIFF
--- a/proto/cachepubsub.proto
+++ b/proto/cachepubsub.proto
@@ -70,13 +70,20 @@ message _SubscriptionRequest {
   // The literal topic name to which you want to subscribe.
   string topic = 2;
 
-  // --> Providing this is not required. <--
-  //
   // If provided, attempt to reconnect to the topic and replay messages starting from
   // the provided sequence number. You may get a discontinuity if some (or all) of the
   // messages are not available.
+  // If provided at 1, you may receive some messages leading up to whatever the
+  // newest message is. The exact amount is unspecified and subject to change.
   // If not provided (or 0), the subscription will begin with the latest messages.
   uint64 resume_at_topic_sequence_number = 3;
+
+  // Determined by the service when a topic is created. This clarifies the intent of
+  // a subscription, and ensures the right messages are sent for a given
+  // `resume_at_topic_sequence_number`.
+  // Include this in your Subscribe() calls when you are reconnecting. The right value
+  // is the last sequence_stamp you received.
+  uint64 sequence_stamp = 4;
 }
 
 // Possible message kinds from a topic. They can be items when they're from you, or
@@ -101,19 +108,20 @@ message _SubscriptionItem {
 
 // Your subscription has yielded an item you previously published. Here it is!
 message _TopicItem {
-  // Topic sequence numbers are **best-effort** and **informational**.
-  // They are not transactional.
-  // They exist:
-  // * to help reconnect to an existing topic while trying to avoid missing items.
-  // * to facilitate richer monitoring and logging.
-  // * to provide a best-effort awareness of stream contiguity, or lack thereof,
-  //   in case you need to know.
-  // You can safely ignore them if none of that matters to you!
+  // Topic sequence numbers give an order of messages per-topic.
+  // All subscribers to a topic will receive messages in the same order, with the same sequence numbers.
   uint64 topic_sequence_number = 1;
+  
   // The value you previously published to this topic.
   _TopicValue value = 2;
+  
   // Authenticated id from Publisher's disposable token
   string publisher_id = 3;
+
+  // Sequence stamps exist to determine which sequence number range a message belongs to. On a topic reset,
+  // the sequence numbers reset and a new sequence_stamp is given.
+  // For a given sequence_stamp, the next message in a topic is topic_sequence_number + 1.
+  uint64 sequence_stamp = 4;
 }
 
 // A value in a topic - published, duplicated and received in a subscription.
@@ -135,6 +143,11 @@ message _Discontinuity {
   uint64 last_topic_sequence = 1;
   // The new topic sequence number after which TopicItems will ostensibly resume.
   uint64 new_topic_sequence = 2;
+  // The new topic sequence_stamp. If you had one before and this one is different, then your topic reset.
+  // If you didn't have one, then this is just telling you what the sequence stamp is expected to be.
+  // If you had one before, and this one is the same, then it's just telling you that you missed some messages
+  // in the topic. Probably your client is consuming messages a little too slowly in this case!
+  uint64 new_sequence_stamp = 3;
 }
 
 // A message from Momento for when we want to reassure clients or frameworks that a
--- a/proto/cachepubsub.proto
+++ b/proto/cachepubsub.proto
@@ -82,8 +82,8 @@ message _SubscriptionRequest {
   // a subscription, and ensures the right messages are sent for a given
   // `resume_at_topic_sequence_number`.
   // Include this in your Subscribe() calls when you are reconnecting. The right value
-  // is the last sequence_stamp you received.
-  uint64 sequence_stamp = 4;
+  // is the last sequence_page you received.
+  uint64 sequence_page = 4;
 }
 
 // Possible message kinds from a topic. They can be items when they're from you, or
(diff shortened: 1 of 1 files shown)


PDXKimani (dismissed):

proto/cachepubsub.proto
```diff
@@ -70,13 +70,20 @@ message _SubscriptionRequest {
...
+  // newest message is. The exact amount is unspecified and subject to change.
   // If not provided (or 0), the subscription will begin with the latest messages.
   uint64 resume_at_topic_sequence_number = 3;
+
+  // Determined by the service when a topic is created. This clarifies the intent of
+  // a subscription, and ensures the right messages are sent for a given
+  // `resume_at_topic_sequence_number`.
+  // Include this in your Subscribe() calls when you are reconnecting. The right value
+  // is the last sequence_stamp you received.
+  uint64 sequence_stamp = 4;
```
anp13, line 86: nit: `topic_sequence_stamp` so that the field naming is similar to `topic_sequence_number`.

proto/cachepubsub.proto
```diff
@@ -135,6 +143,11 @@ message _Discontinuity {
   uint64 last_topic_sequence = 1;
   // The new topic sequence number after which TopicItems will ostensibly resume.
   uint64 new_topic_sequence = 2;
+  // The new topic sequence_stamp. If you had one before and this one is different, then your topic reset.
+  // If you didn't have one, then this is just telling you what the sequence stamp is expected to be.
+  // If you had one before, and this one is the same, then it's just telling you that you missed some messages
+  // in the topic. Probably your client is consuming messages a little too slowly in this case!
+  uint64 new_sequence_stamp = 3;
```
anp13, line 150: nit: `new_topic_sequence_stamp`

proto/cachepubsub.proto
kvcache: I don't really like the idea of mixing the names too closely. Stamp versus number being the only discriminator between the two seems confusing, particularly to someone with English as a second, third, or fourth language. It's already subtle enough haha

anp13: Yea I did have that same thought. If anything I would rather change `sequence` to something else instead of adding `topic` to it.

proto/cachepubsub.proto
```diff
@@ -70,13 +70,20 @@ message _SubscriptionRequest {
   // The literal topic name to which you want to subscribe.
   string topic = 2;
 
-  // --> Providing this is not required. <--
-  //
   // If provided, attempt to reconnect to the topic and replay messages starting from
   // the provided sequence number. You may get a discontinuity if some (or all) of the
   // messages are not available.
+  // If provided at 1, you may receive some messages leading up to whatever the
+  // newest message is. The exact amount is unspecified and subject to change.
```
malandis, line 77: Will topic sequence number `1` ever be sent as a `resume_at_topic_sequence_number`? When would a client provide this number?

proto/cachepubsub.proto
```diff
@@ -70,13 +70,20 @@ message _SubscriptionRequest {
...
   // If not provided (or 0), the subscription will begin with the latest messages.
   uint64 resume_at_topic_sequence_number = 3;
+
+  // Determined by the service when a topic is created. This clarifies the intent of
+  // a subscription, and ensures the right messages are sent for a given
+  // `resume_at_topic_sequence_number`.
+  // Include this in your Subscribe() calls when you are reconnecting. The right value
+  // is the last sequence_stamp you received.
```
malandis, line 85: Or `0` on the initial subscribe?

malandis (commented):
Few clarifying questions on first read, thinking about client user experience

proto/cachepubsub.proto
kvcache: It's used routinely by users of http long polling (ask khawaja, he's a big fan of resume at 1)

proto/cachepubsub.proto
kvcache: "0" and "not included" mean the same thing, so "yes" and "sure, if you like" are equivalent answers to your question I think!

proto/cachepubsub.proto
```diff
@@ -135,6 +143,11 @@ message _Discontinuity {
   uint64 last_topic_sequence = 1;
   // The new topic sequence number after which TopicItems will ostensibly resume.
   uint64 new_topic_sequence = 2;
+  // The new topic sequence_stamp. If you had one before and this one is different, then your topic reset.
```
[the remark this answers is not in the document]
kvcache: yes. With this feature, it will do so more reliably too.

malandis (approved):

kvcache:
@malandis FYI I switched the name here after some discussion. Let me know what you think!

anp13 (dismissed):

proto/cachepubsub.proto
```diff
@@ -118,10 +118,10 @@ message _TopicItem {
   // Authenticated id from Publisher's disposable token
   string publisher_id = 3;
 
-  // Sequence stamps exist to determine which sequence number range a message belongs to. On a topic reset,
-  // the sequence numbers reset and a new sequence_stamp is given.
-  // For a given sequence_stamp, the next message in a topic is topic_sequence_number + 1.
-  uint64 sequence_stamp = 4;
+  // Sequence pages exist to determine which sequence number range a message belongs to. On a topic reset,
+  // the sequence numbers reset and a new sequence_page is given.
+  // For a given sequence_page, the next message in a topic is topic_sequence_number + 1.
```
malandis, line 123: And there is no expectation the sequence pages are sequential, right? If correct, we should document here as the some could assume that from the word `page`.

kvcache: Addressed in https://github.com/momentohq/client-protos/commit/f8491912ef2843561e58ab19b8af0d7f829122ef

kvcache: @malandis is this resolved to your satisfaction?

malandis (approved):

proto/cachepubsub.proto
```diff
@@ -101,19 +108,22 @@ message _SubscriptionItem {
...
 message _TopicItem {
-  // Topic sequence numbers are **best-effort** and **informational**.
-  // They are not transactional.
-  // They exist:
-  // * to help reconnect to an existing topic while trying to avoid missing items.
-  // * to facilitate richer monitoring and logging.
-  // * to provide a best-effort awareness of stream contiguity, or lack thereof,
-  //   in case you need to know.
-  // You can safely ignore them if none of that matters to you!
+  // Topic sequence numbers give an order of messages per-topic.
```
krispraws, line 111: Nit: They give an order per-topic-sequence-page.

krispraws (approved):
LGTM. `page` doesn't quite feel 100% right to me but I don't have a better word to describe a time slice for a topic

danielamiao:
> LGTM. `page` doesn't quite feel 100% right to me but I don't have a better word to describe a time slice for a topic

I don't feel strongly about this either but just curious why switch from `stamp` to `page`? I liked `stamp` better actually, it gave me the right semantics about what it represents

krispraws:
> > LGTM. `page` doesn't quite feel 100% right to me but I don't have a better word to describe a time slice for a topic
> 
> I don't feel strongly about this either but just curious why switch from `stamp` to `page`? I liked `stamp` better actually, it gave me the right semantics about what it represents

I am the opposite - `stamp` did not convey any meaning to me. `instance`, `version`, `page` came closer.

kvcache:
> > > LGTM. `page` doesn't quite feel 100% right to me but I don't have a better word to describe a time slice for a topic
>>>
> > I don't feel strongly about this either but just curious why switch from `stamp` to `page`? I liked `stamp` better actually, it gave me the right semantics about what it represents
>>
> I am the opposite - stamp did not convey any meaning to me. instance, version, page came closer.

It is impossible to please everyone with this name. Requested rationale for the change: Stamps are abstract and have no sense of meaning or order, nor even a notion of change. Page evokes a sense of order and continuity, with a clear signal that "pages turn."
It's not a perfect word, but even the perfect word, "epoch," is not universally liked and itself fails to capture the nonfunctionals. It's a more heady word, while the simplicity of "page" captures the needful. Remember, people are going to be interacting with this via http, so this is an api discussion.

danielamiao (approved):

PR #306, opened 2025-02-03
chore: resolve naming conflicts for js-web sdk protos

nidhidhamnani:


PDXKimani (approved):

javascript-web/generate_protos.sh
```diff
@@ -69,10 +69,16 @@ function generate_proto() {
   echo "Backing up protos dir"
   cp -r ../proto ../proto.bak
 
+  # resolves the naming conflicts between permissionmessages and permissionrules
+  $sed_command -i 's/SuperUserPermissions/SuperUserPermissionSet/g' ../proto/permissionrules.proto
+  $sed_command -i 's/ExplicitPermissions/ExplicitPermissionSet/g' ../proto/permissionrules.proto
+  $sed_command -i 's/SuperUser/SuperUserSet/g' ../proto/permissionrules.proto
```
PDXKimani, line 75: You need to drop this `-i`, actually - see how `$sed_command` is constructed.

PDXKimani (changes requested):

PDXKimani (approved):

javascript-web/generate_protos.sh
nidhidhamnani: Ah thanks, I was doing `sed -i` and then saw the env specific command already exists so missed removing `-i`

PR #310, opened 2025-02-28
feat: add SortedSetUnionStore api

krispraws:
This PR adds a New API `SortedSetUnionStore` that does the equivalent of Redis's `ZUNIONSTORE`
The union of elements in the source sorted sets is computed using the given weights and aggregate function, and the result is stored in the target set on which the api is invoked.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -94,6 +94,9 @@ service Scs {
   rpc SortedSetLength(_SortedSetLengthRequest) returns (_SortedSetLengthResponse) {}
   // Returns number of elements in the sorted set between a given min and max score
   rpc SortedSetLengthByScore(_SortedSetLengthByScoreRequest) returns (_SortedSetLengthByScoreResponse) {}
+  // Computes the union of multiple sorted sets using weights and an aggregate function and stores the result in itself
+  // If the sorted set exists, its value is overwritten.
+  rpc SortedSetUnionStore(_SortedSetUnionStoreRequest) returns (_SortedSetUnionStoreResponse) {}
 }
 
 message _GetRequest {
@@ -1023,5 +1026,37 @@ message _SortedSetLengthByScoreResponse {
   oneof sorted_set {
     _Found found = 1;
     _Missing missing = 2;
+  }
+}
+
+message _SortedSetUnionStoreRequest {
+  enum AggregateFunction {
+    SUM = 0;
+    MIN = 1;
+    MAX = 2;
+  }
+
+  message _Source {
+    bytes set_name = 1;
+    optional float weight = 2;
+  }
+
+  bytes set_name = 1;
+  repeated _Source sources = 2;
+  AggregateFunction aggregate = 3;
+  uint64 ttl_milliseconds = 4;
+  bool refresh_ttl = 5;
+}
+
+message _SortedSetUnionStoreResponse {
+  message _Stored {
+    uint32 length = 1;
+  }
+
+  message _NotStored {}
+
+  oneof sorted_set {
+    _Stored stored = 1;
+    _NotStored not_stored = 2;
   }
 }
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -1054,7 +1054,7 @@ message _SortedSetUnionStoreRequest {
   // The destination set where the result of the union is stored.
   bytes set_name = 1;
 
-  // The sets to compute the union for. Note the desination set is not implicitly included as a source.
+  // The sets to compute the union for. Note that the destination set is not implicitly included as a source.
   // It must be explicitly specified in sources if its current contents need to be included in the union.
   repeated _Source sources = 2;
 
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -94,8 +94,13 @@ service Scs {
   rpc SortedSetLength(_SortedSetLengthRequest) returns (_SortedSetLengthResponse) {}
   // Returns number of elements in the sorted set between a given min and max score
   rpc SortedSetLengthByScore(_SortedSetLengthByScoreRequest) returns (_SortedSetLengthByScoreResponse) {}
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -1025,3 +1028,35 @@ message _SortedSetLengthByScoreResponse {
...
+    AggregateFunction aggregate = 3;
+  }
+
+  bytes set_name = 1;
+  repeated _Source sources = 2;
+  uint64 ttl_milliseconds = 3;
+  bool refresh_ttl = 4;
+}
+
+message _SortedSetUnionStoreResponse {
```
krispraws, line 1051: I could make this simpler and return an uint for the number of elements in the resulting destination set. Which would be 0 if the result was empty and destination was not created.
Thoughts?

proto/cacheclient.proto
```diff
@@ -1025,3 +1028,35 @@ message _SortedSetLengthByScoreResponse {
...
+  enum AggregateFunction {
+    SUM = 0;
+    MIN = 1;
+    MAX = 2;
+  }
+
+  message _Source {
+    bytes set_name = 1;
+    optional float weight = 2;
+    AggregateFunction aggregate = 3;
```
kvcache: Aggregate function is _across_ sources, not per-source. You need to bump this up a level to the Request itself

proto/cacheclient.proto
```diff
@@ -1025,3 +1028,35 @@ message _SortedSetLengthByScoreResponse {
...
+message _SortedSetUnionStoreRequest {
+  enum AggregateFunction {
+    SUM = 0;
+    MIN = 1;
+    MAX = 2;
+  }
+
+  message _Source {
+    bytes set_name = 1;
+    optional float weight = 2;
```
kvcache, line 1041: I think this should not be optional

proto/cacheclient.proto
```diff
@@ -1025,3 +1028,35 @@ message _SortedSetLengthByScoreResponse {
...
+    AggregateFunction aggregate = 3;
+  }
+
+  bytes set_name = 1;
+  repeated _Source sources = 2;
+  uint64 ttl_milliseconds = 3;
+  bool refresh_ttl = 4;
```
kvcache: I do not think this is going to be supported on the stored set.

proto/cacheclient.proto
kvcache: I think this can just return the u64 yes

proto/cacheclient.proto
krispraws: Yep - I fixed that a minute ago. Thanks for the catch

proto/cacheclient.proto
krispraws: What is going to be supported? the ttl and refresh ttl capabilities?

kvcache: ttl only

proto/cacheclient.proto
krispraws: Ok, I'll remove it here and let the SDK handle defaults.

proto/cacheclient.proto
krispraws: All the other SortedSetLength* functions return uint32. It'd be inconsistent to change just this one to u64

krispraws: https://github.com/momentohq/client-protos/pull/310/commits/4d16a5dd8ce4a0cd9e6836374dd600d26d14442c

proto/cacheclient.proto
krispraws: https://github.com/momentohq/client-protos/pull/310/commits/4d16a5dd8ce4a0cd9e6836374dd600d26d14442c

proto/cacheclient.proto
kvcache: 🤷 unioning a few huge sorted sets maybe someday exceeds u32. I don't think the data type matters really, but if you like uint32 then go for it.

proto/cacheclient.proto
krispraws: Removed refresh_ttl in https://github.com/momentohq/client-protos/pull/310/commits/4d16a5dd8ce4a0cd9e6836374dd600d26d14442c

proto/cacheclient.proto
krispraws: You may have to call SortedSetLength on the resulting set also at some point and that returns u32 currently. I can make this u64 if we can live with the inconsistency

kvcache: ah, it doesn't matter much to me right now. I am slowly remembering that you might have some annoying things to deal with in javascript if you use u64 up front (around 2^53 due to ieee754 and js's _interesting_ choices, and dealing with string/number/bigint choices). u32 is fine.

virratanasangpunth (dismissed):

proto/cacheclient.proto
kvcache: I believe uint32 -> uint64 is binary-compatible, so we can widen it in the future if necessary.

proto/cacheclient.proto
```diff
@@ -1025,3 +1028,25 @@ message _SortedSetLengthByScoreResponse {
...
+    MIN = 1;
+    MAX = 2;
+  }
+
+  message _Source {
+    bytes set_name = 1;
+    float weight = 2;
+  }
+
+  bytes set_name = 1;
```
virratanasangpunth, line 1044: nit: can we call it `destination_set_name` for clarity?

krispraws: I think it is implicit that this is the destination by having `Store` in the name of the api and it follows the pattern of other apis

kvcache (approved):
Looks great, thanks

virratanasangpunth (approved):

proto/cacheclient.proto
krispraws: I'll leave this as u32. We can do a wholescale widening in future.

PR #311, opened 2025-03-12
feat: add check-and-set apis based on token

krispraws:
Allow conditionally updating an item based on the value of a token field that is stored alongside the item.
Similar to memcached `cas` and `gets`

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -24,6 +24,8 @@ service Scs {
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc SetBatch (_SetBatchRequest) returns (stream _SetResponse) {}
   rpc SetIf (_SetIfRequest) returns (_SetIfResponse) {}
+  // Conditionally set an item based on the value of a token that is stored along with the item
+  rpc SetIfToken (_SetIfTokenRequest) returns (_SetIfTokenResponse) {}
 
   // Deprecated because we have SetIf - Absent to cover this case.
   rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {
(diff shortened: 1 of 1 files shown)


proto/cacheclient.proto
```diff
@@ -177,6 +180,29 @@ message _SetIfNotExistsResponse {
   message _NotStored { }
 }
 
+// tokens are restricted to 8 bytes
```
krispraws, line 183: Can the Javascript SDK handle uint64 or is it better to model the token as byte arrays and restrict them to 8 bytes in the implementation?

proto/cacheclient.proto
```diff
@@ -177,6 +180,29 @@ message _SetIfNotExistsResponse {
...
+// When setting an item for the first time, previous token is set to 0
+message _SetIfTokenRequest {
```
krispraws, line 185: Alternate version that also allows a conditional delete
```
message _SetIfTokenRequest {
    bytes cache_key = 1;
    uint64 previous_token = 2;
    oneof operation {
        _Replace replace = 3;
        _Delete delete = 4;
    }
    // Overwrite the item with a new value, ttl and token
    message _Replace {
        bytes cache_body = 1;
        uint64 ttl_milliseconds = 2;
        uint64 new_token = 3;
    }
    // Delete the item
    message _Delete {}
}
```

anitarua: I think adding Delete would make the Stored/NotStored responses kind of confusing to deal with. Maybe better to have a separate DeleteIfToken if regular Delete is not sufficient?

proto/cacheclient.proto
```diff
@@ -262,6 +288,21 @@ message _ItemGetTypeResponse {
...
+message _ItemGetTokenResponse {
+  message _Found{
+    uint64 token = 1;
+  }
+  message _Missing{}
+  oneof result {
+    _Found found = 1;
+    _Missing missing = 2;
+  }
+}
```
anitarua, line 304: What is the `Missing.found` case? Does it mean item was found but had no token?

anitarua (commented):
couple of questions

proto/cacheclient.proto
malandis: I have the same question. Do we need to model `key exists and has cas token` vs `key exists but has no cas token` vs `key does not exist`?

krispraws:
I am going to withdraw this since I have made changes to the proposed APIs in https://www.notion.so/momentohq/New-Check-And-Set-Api-similar-to-memcached-1b3feb15c97d80be8b6ccffa0b7f7031?pvs=4#1b6feb15c97d80c28f76ef5862d2efa0
Some comments are already addressed and some are no longer applicable.
I will put up a new PR shortly

PR #312, opened 2025-03-14
feat: new apis to check-and-set an item based on a hash

krispraws:
New version of https://github.com/momentohq/client-protos/pull/311 based on recent discussions.
Now updated to do conditional sets based on value hashes.

### DIFF
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -193,13 +193,13 @@ message _SetIfHashRequest {
     common.PresentAndNotHashEqual present_and_not_hash_equal = 4;
     // Update the item only if it exists and the hash of its current value
     // matches the hash passed in the request    
-    common.HashEqual hash_equal = 5;
+    common.PresentAndHashEqual present_and_hash_equal = 5;
     // Update the item if it exists and the hash of its current value matches
     // the hash passed in the request or create it if the item doesn't exist
     common.AbsentOrHashEqual absent_or_hash_equal = 6;
     // Update the item if it exists and the hash of its current value doesn't
     // match the hash passed in the request or create it if the item doesn't exist
-    common.NotHashEqual not_hash_equal = 7;
+    common.AbsentOrNotHashEqual absent_or_not_hash_equal = 7;
     // Unconditionally update the item if it exists or create it if it doesn't
     common.Unconditional unconditional = 8;
   }
--- a/proto/cacheclient.proto
+++ b/proto/cacheclient.proto
@@ -20,10 +20,15 @@ enum ECacheResult {
 
 service Scs {
   rpc Get (_GetRequest) returns (_GetResponse) {}
+  rpc GetWithHash (_GetWithHashRequest) returns (_GetWithHashResponse) {}
   rpc GetBatch (_GetBatchRequest) returns (stream _GetResponse) {}
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc SetBatch (_SetBatchRequest) returns (stream _SetResponse) {}
   rpc SetIf (_SetIfRequest) returns (_SetIfResponse) {}
+  // Conditionally set the item value based on comparing the hashes computed on the item values
+  // instead of comparing the entire values. This is an alternative to SetIf that saves bandwidth
+  // by removing the need to send the entire value to compare against.
+  rpc SetIfHash (_SetIfHashRequest) returns (_SetIfHashResponse) {}
 
   // Deprecated because we have SetIf - Absent to cover this case.
   rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {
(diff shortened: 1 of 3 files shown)


proto/cacheclient.proto
```diff
@@ -105,13 +109,24 @@ service Scs {
 }
 
 message _GetRequest {
+  message _GetOptions {
+    // Return item token in response
+    bool with_token = 1;
+    // Do not return the item body
+    bool skip_value = 2;
```
krispraws, line 116: This is an optimization we can defer to v2.

proto/cacheclient.proto
```diff
@@ -162,6 +177,26 @@ message _SetIfResponse {
...
+}
+
+message _SetIfTokenResponse {
+  oneof result {
+    _Stored stored = 1;
+    _NotStored not_stored = 2;
+  }
+  // Stored with a new token
+  message _Stored {
+    bytes token = 1;
```
krispraws, line 194: Returning new token in the response is a convenience - we could also force users to call ItemGetToken or Get instead.

nidhidhamnani (approved):

proto/cacheclient.proto
krispraws: In order to keep things as simple as possible for the MVP, I am going to skip returning the new token in the Set response. For an application where concurrent updates happen regularly, users will anyway have to call GetToken to get the latest token. If anyone has objections, let me know.

proto/cacheclient.proto
krispraws: Any opinions on adding `skip_value` as an option to Get to only get the token vs adding a new api ItemGetToken? I prefer the former.

krispraws: Correct. I can also make the `options` field optional if that helps

anitarua: yeah I think making it optional would be good

krispraws: Done in f5358c3a7f3a407946b07918deda993dbb83aa5f

proto/cacheclient.proto
```diff
@@ -262,6 +294,21 @@ message _ItemGetTypeResponse {
...
+message _ItemGetTokenResponse {
+  message _Found{
+    bytes token = 1;
+  }
+  message _Missing{}
+  oneof result {
+    _Found found = 1;
+    _Missing missing = 2;
+  }
+}
```
anitarua, line 310: I'm still unclear on what the Missing cases are. 

- Missing.Found means the token was missing but item was found?
- Missing.Missing means item was missing?

Because of this, I think I'm in favor of removing ItemGetToken and using the optional Get metadata instead. At least then it's clear that if Get returns Missing, the entire item + token was missing

krispraws: The token's lifetime is tied to the item's, so there is no question of one being missing when the other is not.

krispraws: If an item exists, it will always have a token whose value is generated by Momento.

anitarua: Okay, thanks for clarifying! Also I double checked and I guess the sdks don't parse the Missing subcases anyway for `ItemGet` calls.

Getting token via either method seems fine. From the sdk side, we could add either or both:
- `CacheGetHit.token()` to return `string | undefined`
- `ItemGetTokenResponse` -- `Hit.token()` will return a string, `Miss` will return nothing. More definitive than `string | undefined`

proto/cacheclient.proto
```diff
@@ -162,6 +183,40 @@ message _SetIfResponse {
...
+    common.PresentAndNotHashEqual present_and_not_hash_equal = 4;
+    // Update the item only if it exists and the hash of its current value
+    // matches the hash passed in the request    
+    common.HashEqual hash_equal = 5;
+    // Update the item if it exists and the hash of its current value matches
+    // the hash passed in the request or create it if the item doesn't exist
+    common.AbsentOrHashEqual absent_or_hash_equal = 6;
+    // Update the item if it exists and the hash of its current value doesn't
+    // match the hash passed in the request or create it if the item doesn't exist
+    common.NotHashEqual not_hash_equal = 7;
```
anitarua, line 202: could these be renamed for clarity?
```suggestion
    common.PresentAndNotHashEqual present_and_not_hash_equal = 4;
    // Update the item only if it exists and the hash of its current value
    // matches the hash passed in the request    
    common.PresentAndHashEqual hash_equal = 5;
    // Update the item if it exists and the hash of its current value matches
    // the hash passed in the request or create it if the item doesn't exist
    common.AbsentOrHashEqual absent_or_hash_equal = 6;
    // Update the item if it exists and the hash of its current value doesn't
    // match the hash passed in the request or create it if the item doesn't exist
    common.AbsentOrNotHashEqual not_hash_equal = 7;
```

krispraws: Done in https://github.com/momentohq/client-protos/pull/312/commits/6c62c01a19293b188499ae5dc3de9522a155695f

anitarua (approved):

PR #315, opened 2025-03-27
chore: specify javascript grpc dependencies as peer dependencies

malandis:
Because the JavaScript SDKs must also have the grpc
library (`@grpc/grpc-js` for node; `grpc-web` for web) installed, the
proto libraries here should declare grpc as a peer dependency, not
production dependency.

In the previous form as production dependencies, both the SDK and the
protos libraries install grpc as a dependency. Because we create grpc symbols
in the SDK to pass to objects from the protos library, and because the grpc library
performs an instanceof check on them, those symbols *must* be loaded from the
same copy of the grpc library. That is, the node module loader must be using the identical
(as in same files on the filesystem) in order for the instanceof check to pass.
In some situations, deduplication (same grpc library and same version) fails,
which causes the instanceof check to fail.

By mandating grpc as a peer dependency, we avoid this
altogether. There will be one copy of grpc to load between the SDK and
the protos library: the one the SDK installs.

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -15,6 +15,7 @@
   "author": "",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@grpc/grpc-js": "1.13.1",
     "@tsconfig/node16": "1.0.2",
     "@types/google-protobuf": "3.15.6",
     "@types/node": "16.10.3",
@@ -23,8 +24,10 @@
     "typescript": "4.9.5"
   },
   "dependencies": {
-    "@grpc/grpc-js": "1.13.1",
     "google-protobuf": "3.21.2"
+  },
+  "peerDependencies": {
+    "@grpc/grpc-js": "1.13.1"
   },
   "files": [
     "dist"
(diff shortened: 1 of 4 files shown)


pgautier404 (approved):

javascript/package.json
```diff
@@ -23,9 +24,11 @@
     "typescript": "4.9.5"
   },
   "dependencies": {
-    "@grpc/grpc-js": "1.13.1",
     "google-protobuf": "3.21.2"
   },
+  "peerDependencies": {
+    "@grpc/grpc-js": "1.13.1"
```
rishtigupta, line 30: This version  would be the minimum version that is required to be installed by the application?

malandis: Exact version the SDK would need. We're going to stay with pinned versions for now.

rishtigupta: If someone developing the SDK doesn’t notice the peer dependency block, will they get an error that clearly tells them which specific version they need to install? Apologies if I’m reading too much into this.

malandis: The SDK will have `@grpc/grpc-js` as a `dependency` with a version that matches this (the peer dependency resolution would ensure this when running `npm i @gomomento/generated-types`). That said, it's already listed as a dependency in the SDK so a developer just needs to run `npm install` and it'll just work.

rishtigupta (approved):
:shipit:

PR #317, opened 2025-03-27
chore: javascript(-web) should specify protobuf as a peer dependency

malandis:
Similar to #315, protobuf should be a peer dependency in the
JavaScript projects. Since the SDKs install protobuf alongside grpc,
they are vulnerable to the same deduplication problem as we had with
grpc.


rishtigupta (approved):
:shipit:

